PR #1, opened 2020-08-12
add rke2 charts

MonzElmasry:
This pr moves rke2-charts in this repo, Rename charts to have rke2- prefix and regenerate patches 
issue :https://github.com/rancher/rke2/issues/161

### DIFF
--- a/packages/rke2-ingress-nginx/package.yaml
+++ b/packages/rke2-ingress-nginx/package.yaml
@@ -1,0 +1,2 @@
+url: https://kubernetes-charts.storage.googleapis.com/nginx-ingress-1.36.3.tgz
+packageVersion: 00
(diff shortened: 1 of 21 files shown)


packages/rke2-ingress-nginx/package.yaml
```diff
@@ -0,0 +1,2 @@
+url: https://kubernetes-charts.storage.googleapis.com/nginx-ingress-1.36.3.tgz
```
StrongMonkey, line 1: If you rename this, you might have to make sure the chart.yaml's name is also renamed to rke2-ingress-nginx

packages/rke2-metrics-server/package.yaml
```diff
@@ -0,0 +1,2 @@
+url: https://kubernetes-charts.storage.googleapis.com/metrics-server-2.11.1.tgz
```
StrongMonkey, line 1: same. Make sure the chart.yaml is also renamed.

MonzElmasry: this isnt renamed.

PR #3, opened 2020-08-14
change k8s-app to kube-dns

MonzElmasry:
This pr changes k8s-app from `coredns` to `kube-dns` for promethus to be able to detect coredns service, needed for rke2 cluster monitoring 
issue: https://github.com/rancher/rke2/issues/33

### DIFF
--- a/packages/rke2-coredns/rke2-coredns.patch
+++ b/packages/rke2-coredns/rke2-coredns.patch
@@ -10,9 +10,78 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rke2-coredns/charts-original/Chart.yam
  sources:
  - https://github.com/coredns/coredns
  version: 1.10.1
+diff -x '*.tgz' -x '*.lock' -uNr packages/rke2-coredns/charts-original/templates/clusterrole-autoscaler.yaml packages/rke2-coredns/charts/templates/clusterrole-autoscaler.yaml
+--- packages/rke2-coredns/charts-original/templates/clusterrole-autoscaler.yaml
++++ packages/rke2-coredns/charts/templates/clusterrole-autoscaler.yaml
+@@ -9,7 +9,7 @@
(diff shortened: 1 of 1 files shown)


packages/rke2-coredns/rke2-coredns.patch
```diff
@@ -10,9 +10,78 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rke2-coredns/charts-original/Chart.yam
...
  version: 1.10.1
+diff -x '*.tgz' -x '*.lock' -uNr packages/rke2-coredns/charts-original/templates/clusterrole-autoscaler.yaml packages/rke2-coredns/charts/templates/clusterrole-autoscaler.yaml
+--- packages/rke2-coredns/charts-original/templates/clusterrole-autoscaler.yaml
++++ packages/rke2-coredns/charts/templates/clusterrole-autoscaler.yaml
+@@ -9,7 +9,7 @@
+     app.kubernetes.io/instance: {{ .Release.Name | quote }}
+     helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+     {{- if .Values.isClusterService }}
+-    k8s-app: {{ .Chart.Name }}-autoscaler
++    k8s-app: {{ .Values.k8sApp }}-autoscaler
```
aiyengar2, line 21: If `.Values.k8sApp` is undefined, what happens? At the moment, it seems like this would be labeled `-autoscaler`.

Ideally, we should either use [`required`](https://helm.sh/docs/howto/charts_tips_and_tricks/) to ensure that `.Values.k8sApp` is set here or [`default`](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/) to ensure that we fall back to `.Chart.Name` if `.Values.k8sApp` does not exist.

This comment should apply in several places across the PR.

PR #26, opened 2020-10-22
Add cilium v1.8.4 chart

manuelbuil:
As explained in the issue: https://github.com/rancher/rke2/issues/490, it would be great to have Cilium available as CNI.

This patch includes (or hopes to include) the required files to generate the cilium helm chart to be consumed by rke2.


mrostecki (approved):

manuelbuil:
This approach is better ==> https://github.com/rancher/rke2-charts/pull/46. Therefore, this PR gets cancelled

PR #29, opened 2020-12-07
Update coredns to create a service account

briandowns:
Updates the CoreDNS chart to require CoreDNS to create a service account for itself. This is required for the CIS 1.5 5.1.5 control.

### DIFF
--- a/packages/rke2-coredns/rke2-coredns.patch
+++ b/packages/rke2-coredns/rke2-coredns.patch
@@ -330,6 +330,19 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rke2-coredns/charts-original/values.ya
    pullPolicy: IfNotPresent
  
  replicaCount: 1
+@@ -34,10 +34,10 @@
+     prometheus.io/port: "9153"
+ 
+ serviceAccount:
+-  create: false
++  create: true
+   # The name of the ServiceAccount to use
+   # If not set and create is true, a name is generated using the fullname template
+-  name:
++  name: coredns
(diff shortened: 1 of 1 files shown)


MonzElmasry (approved):
LGTM

dweomer (approved):

packages/rke2-coredns/rke2-coredns.patch
```diff
@@ -330,6 +330,19 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rke2-coredns/charts-original/values.ya
...
+@@ -34,10 +34,10 @@
+     prometheus.io/port: "9153"
+ 
+ serviceAccount:
+-  create: false
++  create: true
+   # The name of the ServiceAccount to use
+   # If not set and create is true, a name is generated using the fullname template
+-  name:
++  name: coredns
```
brandond, line 342: Is there any reason to hardcode the name instead of just letting the chart generate it? If we're hardcoding this here instead of letting the chart generate a name, it looks we also have to hardcode the name at https://github.com/rancher/rke2-charts/blob/main/charts/rke2-coredns/values.yaml#L49

For CIS, do we want to also enable PSP for CoreDNS here: https://github.com/rancher/rke2-charts/blob/main/charts/rke2-coredns/values.yaml#L46

briandowns: I opted to hardcode it with the assumption that the old name would be left behind and we'd start littering names on each upgrade.

PR #30, opened 2020-12-10
coredns: v1.7.1

dweomer:
Part of rancher/rke2#345

### DIFF
--- a/packages/rke2-coredns/rke2-coredns.patch
+++ b/packages/rke2-coredns/rke2-coredns.patch
@@ -1,15 +1,24 @@
 diff -x '*.tgz' -x '*.lock' -uNr packages/rke2-coredns/charts-original/Chart.yaml packages/rke2-coredns/charts/Chart.yaml
 --- packages/rke2-coredns/charts-original/Chart.yaml
 +++ packages/rke2-coredns/charts/Chart.yaml
-@@ -17,7 +17,7 @@
-   name: andor44
- - email: manuel@rueg.eu
-   name: mrueg
+@@ -1,7 +1,6 @@
+ apiVersion: v1
+ appVersion: 1.7.1
+-deprecated: true
+-description: DEPRECATED CoreDNS is a DNS server that chains plugins and provides Kubernetes
++description: CoreDNS is a DNS server that chains plugins and provides Kubernetes
+   DNS Services
+ home: https://coredns.io
+ icon: https://coredns.io/images/CoreDNS_Colour_Horizontal.png
+@@ -9,7 +8,7 @@
+ - coredns
+ - dns
+ - kubedns
 -name: coredns
 +name: rke2-coredns
  sources:
  - https://github.com/coredns/coredns
- version: 1.10.1
+ version: 1.13.8
 diff -x '*.tgz' -x '*.lock' -uNr packages/rke2-coredns/charts-original/templates/clusterrole-autoscaler.yaml packages/rke2-coredns/charts/templates/clusterrole-autoscaler.yaml
 --- packages/rke2-coredns/charts-original/templates/clusterrole-autoscaler.yaml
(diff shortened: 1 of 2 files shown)


packages/rke2-coredns/rke2-coredns.patch
```diff
@@ -167,6 +176,150 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rke2-coredns/charts-original/templates
...
++            {{ end }}
++{{- range .Values.extraSecrets }}
++        - name: {{ .name }}
++          secret:
++            secretName: {{ .name }}
++            defaultMode: 400
++{{- end }}
++{{- if .Values.extraVolumes }}
++{{ toYaml .Values.extraVolumes | indent 8 }}
++{{- end }}
```
dweomer, line 322: This is the part that worries me: its replacing the deployment.yaml, I think?

packages/rke2-coredns/rke2-coredns.patch
```diff
@@ -314,24 +467,68 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rke2-coredns/charts-original/templates
...
-@@ -3,8 +3,8 @@
+@@ -3,7 +3,7 @@
  # Declare variables to be passed into your templates.
  
  image:
 -  repository: coredns/coredns
--  tag: "1.6.9"
 +  repository: rancher/hardened-coredns
-+  tag: "v1.6.9"
+   tag: "1.7.1"
```
brandond, line 527: The actual tag here seems to be `v1.7.1`
https://hub.docker.com/layers/rancher/hardened-coredns/v1.7.1/images/sha256-2f943b97ae07055f177f9c89a429bf8a2b0afd85a4f85abc51def84773fef814
```suggestion
   tag: "v1.7.1"
```

brandond (changes requested):
I can't make heads or tails of what this is actually deploying for a corefile, but there is a change necessary between v1.6.9 and v1.7.1, see
https://github.com/k3s-io/k3s/commit/cd27c6fcbe5fe266b20f682374f09cf52661ceeb#diff-1c144ada012fb4c7c809e2fbddfac3fa2acbae57d85ebbce654c94944f667252L61

packages/rke2-coredns/rke2-coredns.patch
dweomer: Yes, but, https://hub.docker.com/layers/rancher/hardened-coredns/v1.6.9/images/sha256-97f077ce82448672f87a98b54a8da5ebd46ca01aeb4be6e59990399b470a0865?context=explore

packages/rke2-coredns/rke2-coredns.patch
```diff
@@ -314,24 +467,68 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rke2-coredns/charts-original/templates
...
    pullPolicy: IfNotPresent
  
- replicaCount: 1
-@@ -34,10 +34,10 @@
-     prometheus.io/port: "9153"
+@@ -66,10 +66,10 @@
+   annotations: {}
  
  serviceAccount:
```
briandowns, line 533: Do we want to flip this to true and include the name field like we currently have?

packages/rke2-coredns/rke2-coredns.patch
dweomer: Oh, I see what i did. :bonk:

packages/rke2-coredns/rke2-coredns.patch
brandond: I can't tell what this patch is actually doing - are we dropping the replicaCount, and Prometheus scrape port annotations?

dweomer: no, its just the change is no longer anchored to that text

dweomer: @briandowns create service account is being set to true via the values file

PR #37, opened 2020-12-17
Honor data-dir when templating kubeconfig volumeMount and path

brandond:
Turns out the kubeconfig has a bunch of absolute paths in it, so we need to template both the path to the file AND the volumeMount path when changed by data-dir.


galal-hussein (approved):
LGTM

briandowns (approved):

PR #38, opened 2021-01-08
Update stable charts repo URL and set priorityClassName

brandond:
* Update stable helm chart repo URL
* Set priorityClassName for coredns and metrics-server

Related to rancher/rke2#585

Note that some of the patch sections are reordered due to different versions of `patch` on linux and OS X hosts. There doesn't seem to be a good fix for this, other than having everyone who touches these files agree on what version to use.


briandowns (approved):

galal-hussein (approved):
LGTM

PR #40, opened 2021-01-12
Fix CoreDNS CriticalAddonsOnly toleration

brandond:
The scheduler annotation has been unsupported and ineffective since Kubernetes 1.6 so this is long overdue.

Related to rancher/rke2#585 rancher/rke2#487


briandowns (approved):

dweomer (approved):
diffs od diffs will never not trip me up

PR #42, opened 2021-01-13
Update Kube-Proxy

briandowns:


MonzElmasry (approved):

brandond (approved):

briandowns:
> This looks fine but wow is the workflow for maintain multiple branches annoying.

I agree. After this release is complete, I'd like to think of a way to smooth this out some.

brandond:
Yeah I don't know if we need to start branching or what, but I ran into the same problem with the CoreDNS chart. The whole workflow feels very broken.

briandowns:
Branching was my first though however I'm curious what others are doing in this scenario. I can't imagine we're the first to hit this.

PR #45, opened 2021-01-25
Remove kube-proxy default featureGates

erikwilson:


briandowns (approved):

dweomer (approved):

packages/rke2-kube-proxy/charts/values.yaml
```diff
@@ -39,85 +39,6 @@ detectLocalMode:
 
 # A set of key=value pairs that describe feature gates for alpha/experimental features:
 featureGates:
```
brandond, line 41: I think you will need to set this to an empty map otherwise you will get an error when the chart tries to iterate over a nil value.

```suggestion
featureGates: {}
```

brandond (changes requested):

PR #46, opened 2021-02-09
Add cilium chart

jcaamano:
The chart is organized in subcharts to clearly delimit supported vs
non-supported customization options, as follows:

- The main rke2-cilium chart which only supplies supported customization
  options through its values.yaml.
- Cilium upstream chart is pulled in as a subchart dependency and is
  patched to support a system default registry as a global variable.
- A rke2-cilium-hard-defaults subchart which supplies cilium options
  that change from upstream defaults for which we don't intend to
  support any customization.

All cilium options are scoped & accessible, for example:
`helm install rke2-cilium --set cilium.preflight.enabled=true`

Signed-off-by: Jaime Caamaño Ruiz <jcaamano@suse.com>


packages/rke2-cilium/charts/requirements.yaml
```diff
@@ -0,0 +1,8 @@
+dependencies: # A list of the chart requirements (optional)
+  - name: cilium
+    repository: https://helm.cilium.io
```
brandond, line 3: I think we need to mirror the subcharts, rather than leaving them as an external reference.

packages/rke2-cilium/package.yaml
```diff
@@ -0,0 +1 @@
+packageVersion: 01
```
brandond, line 1: I'm still a little shaky on the new process, but does this need a URL line in order to specify what the upstream chart that the patch is being applied to?

packages/rke2-cilium/charts/requirements.yaml
jcaamano: It's doing it for us:
https://github.com/rancher/rke2-charts/blob/b40561ff50a916ea201798c14904cd7a233c0408/scripts/prepare-subcharts#L81

And documented as such:
https://github.com/rancher/rke2-charts/blob/b40561ff50a916ea201798c14904cd7a233c0408/scripts/prepare-subcharts#L14

Although I am not completly convinced this is not a happy coincidence.

packages/rke2-cilium/package.yaml
jcaamano: It's getting pulled automatically due to being a requirement, yes, as seen above. If it were not a subchart and we would only like to have a parent chart based of a pulled chart then we would need an url here. The url is only strictly check to perform a download that we don't need in this case:

https://github.com/rancher/rke2-charts/blob/b40561ff50a916ea201798c14904cd7a233c0408/scripts/prepare#L11-L34

But `package.yaml` with `packageVersion` is needed to apply the patch and track different versions of the patch itself.

jcaamano:
Checking on the generated chart on my fork produces intended results:

https://github.com/jcaamano/rke2-charts/raw/main/assets/rke2-cilium/rke2-cilium-1.9.401.tgz

brandond (approved):

davidnuzik:
@jcaamano is this unblocked now?

jcaamano:
> @jcaamano is this unblocked now?

@davidnuzik yes, this should be unblocked now

manuelbuil:
Tested in rke2. LGTM!

mrostecki (approved):

PR #52, opened 2021-02-26
[Source] Migrate to charts-build-scripts

aiyengar2:
This PR migrates rke2-charts to using the new charts-build-scripts.

Before merging this PR, the following actions need to be taken:
- [x] Generate a diff of charts that are currently in main against charts that will be generated by this repository (will be stored in external repo)
- [x] Fixup package naming scheme for multiple release lines when applicable

cc: @dweomer

PR #53, opened 2021-02-26
[Live] Migrate to charts-build-scripts

aiyengar2:
This PR migrates rke2-charts to using the new charts-build-scripts.

No changes were introduced to the assets/ (except removing assets/index.yaml) or index.yaml, but merging this PR should be blocked on fixing the following dependencies on the charts/ directory on RKE2's side:
- [Dockerfile](https://github.com/rancher/rke2/blob/master/Dockerfile#L125-L135)
- [build-chart.sh](https://github.com/rancher/rke2/blob/master/charts/build-chart.sh#L5-L20)

Instead of relying on absolute paths, it should reference the index.yaml, which has the source of truth for the real directory that contains the tgz file for a given chart, independent of the package that the chart is contained within.

cc: @dweomer


aiyengar2:
#### Why did you create 2 commits and then revert them?

The last 4 commits were created and reverted in order to ensure the `make sync` workflow works based on the Source branch that will be merged in https://github.com/rancher/rke2-charts/pull/52.

This seems to be working fine, so this PR seems to be good to go to migrate to the new scripts.

#### Are we ready to go?

Yes. No more changes should be necessary from my end for this PR, but this PR must be merged in before https://github.com/rancher/rke2-charts/pull/52, once it has been approved.

_config.yml
```diff
@@ -0,0 +1 @@
+exclude: [charts]
```
aiyengar2, line 1: FYI, this is something we included to force Jekyll (which is what GitHub Pages uses to render your site) to drop the charts/ directory from being served. This was done because Jekyll tries to parse templates the exact same way that Helm does with spacebars (e.g. `{{ value }}`), so we encountered build issues in the partner-charts by exposing this field re: https://github.com/rancher/rancher/issues/31021

dweomer: an in-line comment in the source would be a great place to drop this reminder:
```
prevent jekyll form processing charts templates as if they were jekyll templates
```

dweomer (approved):

aiyengar2:
Merge this in before merging https://github.com/rancher/rke2-charts/pull/52.

erikwilson (approved):

galal-hussein (approved):

PR #69, opened 2021-04-19
Change the default ipam mode of cilium

manuelbuil:
Cilium's default ipam mode does not honor the pod CIDR set in kube-controller-manager.
The benefits of that mode are not relevant when deploying with rke2

Fixes issue rke2/891


jcaamano:
Would this be more appropriate in `packages/rke2-cilium/generated-changes/dependencies/rke2-cilium-hard-defaults/charts/values.yaml` given that we are not interested in users customizing this setting into any other value?

jcaamano (approved):

PR #71, opened 2021-04-26
Add ability to configure calico failsafe ports

jonaz:
Made FailsafeOutboundHostPorts and FailsafeInboundHostPorts configurable
to allow users to have custom network policy for example SSH (22)

Previously failsafe rules would always have opened 22 even if user had
custom GlobalNetworkPolicy applied on the node.

Fixes https://github.com/rancher/rke2/issues/921

### DIFF
--- a/packages/rke2-canal/charts/values.yaml
+++ b/packages/rke2-canal/charts/values.yaml
@@ -71,6 +71,10 @@ calico:
   typhaServiceName: none
   # Kubelet flex-volume-plugin-dir
   flexVolumePluginDir: /var/lib/kubelet/volumeplugins
+  # calico inbound failsafe ports. Empty string means defaults. Use 'none' to disable failsafe if you have your own rules.
+  felixFailsafeInboundHostPorts: ""
+  # calico outbound failsafe ports. Empty string means defaults. Use 'none' to disable failsafe if you have your own rules.
+  felixFailsafeOutboundHostPorts: ""
 
 global:
   systemDefaultRegistry: ""
--- a/packages/rke2-canal/charts/templates/daemonset.yaml
+++ b/packages/rke2-canal/charts/templates/daemonset.yaml
@@ -154,6 +154,10 @@ spec:
               value: {{ .Values.calico.felixPrometheusMetricsEnabled | quote }}
             - name: FELIX_XDPENABLED
               value: {{ .Values.calico.felixXDPEnabled | quote }}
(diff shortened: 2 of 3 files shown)


packages/rke2-canal/charts/values.yaml
```diff
@@ -71,6 +71,10 @@ calico:
   typhaServiceName: none
   # Kubelet flex-volume-plugin-dir
   flexVolumePluginDir: /var/lib/kubelet/volumeplugins
+  # calico inbound failsafe ports. Empty string means defaults. Use 'none' to disable failsafe if you have your own rules.
+  FailsafeInboundHostPorts: ""
```
Oats87, line 75: I believe this should probably be prefixed with a `felix` and formatted in `camelCase` to keep with chart convention.

packages/rke2-canal/charts/values.yaml
```diff
@@ -71,6 +71,10 @@ calico:
...
+  # calico outbound failsafe ports. Empty string means defaults. Use 'none' to disable failsafe if you have your own rules.
+  FailsafeOutboundHostPorts: ""
```
Oats87, line 77: I believe this should probably be prefixed with a `felix` and formatted in `camelCase` to keep with chart convention.

Oats87 (changes requested):

packages/rke2-canal/charts/values.yaml
jonaz: Fixed

packages/rke2-canal/charts/values.yaml
jonaz: Fixed

Oats87 (approved):
LGTM

brandond (approved):

jonaz:
@brandond do i need to do anything else? It says: `1 workflow awaiting approval, First-time contributors need a maintainer to approve running workflows. Learn more.
`

brandond:
You need to also bump the version at https://github.com/rancher/rke2-charts/blob/main-source/packages/rke2-canal/package.yaml#L2

jonaz:
> You need to also bump the version at
> 
> https://github.com/rancher/rke2-charts/blob/1b4dda478d884b21ee98637ea5e60673bfb3fe4c/packages/rke2-canal/package.yaml#L2

Just bumped it to 03. Hope that was correct :)

PR #72, opened 2021-04-26
Add multus & sr-iov charts

jcaamano:
Adds charts for sriov and multus.

Details:

* Multus chart, on deployment installs number of cnis, not only multus
  itself but also other cnis ussually combined with it, like host-device
  and macvlan.
* Multus includes a CRD, NetworkAttachmentDefinition, which is the
  means to attach interfaces to pods by specifying a delegated cni for
  each.
* Multus replaces the existing cni with itself, and sets up a default
  NetworkAttachmentDefinition with that existing cni that will setup the
  first interface of the pod just as if multus was not there.
* The NetworkAttachmentDefinition CRD is included along other multus
  templates in the chart and no specific chart is setup for the CRD.
  This CRD is no consumed on deployment and is unlikely to change
  frequently.
* Sriov depends on NetworkAttachmentDefinition CRD but no dependency is
  set between Multus and Sriov. Sriov charts checks on the presence of
  NetworkAttachmentDefinition CRD via capabilities and fails deployment
  if unavailable.
* Sriov includes a number of CRDs which are setup as separate chart.
* Sriov operator depends on certificates to be able to deploy its
  webhooks. The charts depends and checks for the presence of
  cert-manager to generate these certificates. Optionally, via chart
  value, cert-manager integration can be disabled in which case the
  chart will generate certificates with 1 year validity as a means to
  quickly test or trial, but not the intended setup for production
  clusters.

Signed-off-by: Jaime Caamaño Ruiz <jcaamano@suse.com>

### DIFF
--- a/packages/rke2-multus/charts/Chart.yaml
+++ b/packages/rke2-multus/charts/Chart.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+name: rke2-multus
+description: Multus CNI enables attaching multiple network interfaces to pods in Kubernetes.
+version: v3.7.1-build20210416
+appVersion: v3.7.1
+home: https://github.com/k8snetworkplumbingwg/multus-cni
+icon: https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/master/doc/images/Multus.png
+sources:
+  - https://github.com/k8snetworkplumbingwg/multus-cni
+maintainers:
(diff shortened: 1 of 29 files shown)


manuelbuil (approved):
Tested and works

mrostecki (approved):

manuelbuil (approved):

PR #81, opened 2021-05-12
Honor .Values.global.clusterCIDR in rke2-canal

brandond:
@ibuildthecloud pointed out that the chart does need to use this passthrough cluster setting

### DIFF
--- a/packages/rke2-canal/charts/templates/config.yaml
+++ b/packages/rke2-canal/charts/templates/config.yaml
@@ -60,7 +60,7 @@ data:
   # Flannel network configuration. Mounted into the flannel container.
   net-conf.json: |
     {
-      "Network": {{ .Values.podCidr | quote }},
+      "Network": {{ .coalesce .Values.global.clusterCIDR Values.podCidr | quote }},
       "Backend": {
         "Type": {{ .Values.flannel.backend | quote }}
       }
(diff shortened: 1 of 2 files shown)


briandowns (approved):

packages/rke2-canal/charts/templates/config.yaml
```diff
@@ -60,7 +60,7 @@ data:
   # Flannel network configuration. Mounted into the flannel container.
   net-conf.json: |
     {
-      "Network": {{ .Values.podCidr | quote }},
+      "Network": {{ .coalesce .Values.global.clusterCIDR Values.podCidr | quote }},
```
manuelbuil, line 63: Is this needed? I thought canal was using the kubernetes nodes CIDR as IPAM?

manuelbuil: I have just realized that Values.podCidr has a default value of "10.42.0.0/16". I guess using the kubernetes nodes CIDR as IPAM only happens if this value wouldn't exist, that's why this is needed, right?

brandond: Yes. I thought the same thing, but @ibuildthecloud and @Oats87 clarified that this does need to be kept in sync with the cluster setting.

dweomer (approved):

manuelbuil (approved):

PR #85, opened 2021-05-13
rke2-canal: fix use of coalesce

dweomer:
Signed-off-by: Jacob Blain Christen <jacob@rancher.com>


dweomer:
Supporting fix for rancher/rke2#990

PR #86, opened 2021-05-19
Flannel configure backend port

jzuijlek:
Allow changing of the backend port.


Cuisayi:
We also want this feature. 
Why not merge it for so long?

amomra:
Hi everyone

In our case, this is needed because the default port (8472) conflicts with our VMWare NSX infrastructure. The intracluster communication (between nodes) wasn't working because Flannel was configured to use the 8472 port. After we changed the port to 4789 the communication between nodes started working. Since the newer versions of NSX changed the default port to 4789 (https://kb.vmware.com/s/article/2149996), this feature will allow to install RKE2 in every VMWare environment.

rbrtbnfgl:
Sorry for the delay on this. I added this feature on my PR #307

PR #88, opened 2021-05-20
Revert "Fix bug when deploying with Cilium"

manuelbuil:
This reverts commit 730b1f6398c591fbaf1ca87c96d125310b0a4698.

There is no need to update Cilium because the problem was fixed by ==> https://github.com/k3s-io/helm-controller/pull/98

### DIFF
--- a/packages/cilium/package.yaml
+++ b/packages/cilium/package.yaml
@@ -1,5 +1,5 @@
 url: https://helm.cilium.io/cilium-1.9.6.tgz
-packageVersion: 04
+packageVersion: 05
 releaseCandidateVersion: 00
 # This package is meant to be consumed as a subchart of another package,
 # not directly.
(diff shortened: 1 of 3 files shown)


brandond (approved):

packages/cilium/package.yaml
```diff
@@ -1,5 +1,5 @@
 url: https://helm.cilium.io/cilium-1.9.6.tgz
-packageVersion: 04
```
brandond, line 2: I don't think you can revert the package versions - the versions always need to move forward, even if you're reverting a change to the chart.

manuelbuil: ok, let me change that

brandond (approved):

manuelbuil:
Even though we moved the helm chart, we don't need to update the rke2 though, so I'll not do it. Unless you have a different opinion

PR #94, opened 2021-05-21
[release-1.20] Do not quote kube-proxy config values of type bool

brandond:
For https://github.com/rancher/rke2/issues/1010#issuecomment-845625976

Validated against schema from
https://github.com/kubernetes/kube-proxy/blob/release-1.19/config/v1alpha1/types.go

### DIFF
--- a/packages/rke2-kube-proxy/charts/Chart.yaml
+++ b/packages/rke2-kube-proxy/charts/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: rke2-kube-proxy
 description: Install Kube Proxy.
-version: v1.21.1-build20210520
-appVersion: v1.21.1
+version: v1.20.7-build20210520
+appVersion: v1.20.7
 keywords:
   - kube-proxy
 sources:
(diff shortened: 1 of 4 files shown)


packages/rke2-kube-proxy/charts/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: rke2-kube-proxy
 description: Install Kube Proxy.
-version: v1.21.1-build20210520
-appVersion: v1.21.1
+version: v1.20.7-build20210520
+appVersion: v1.20.7
```
manuelbuil, line 5: Wondering how we control in rke2-charts the release. Will this generate an extra release for the kube-proxy chart that we can consume with version v1.20.7-build20210520? Or will it remove the v1.21.1-build20210520?

brandond: The package version gets appended to this, so it will create a new chart version: v1.20.7-build2021052005

erikwilson (approved):

briandowns (approved):

PR #95, opened 2021-05-21
[release-1.19] Do not quote kube-proxy config values of type bool

brandond:
For https://github.com/rancher/rke2/issues/1010#issuecomment-845625976

Validated against schema from
https://github.com/kubernetes/kube-proxy/blob/release-1.19/config/v1alpha1/types.go

### DIFF
--- a/packages/rke2-kube-proxy/charts/Chart.yaml
+++ b/packages/rke2-kube-proxy/charts/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: rke2-kube-proxy
 description: Install Kube Proxy.
-version: v1.21.1-build20210520
-appVersion: v1.21.1
+version: v1.19.11-build20210520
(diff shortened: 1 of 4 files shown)


packages/rke2-kube-proxy/charts/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: rke2-kube-proxy
 description: Install Kube Proxy.
-version: v1.21.1-build20210520
-appVersion: v1.21.1
+version: v1.18.19-build20210520
```
erikwilson, line 4: Should this be 1.19?

brandond: Yes, thank you :(

briandowns (approved):

erikwilson (approved):

PR #98, opened 2021-05-26
Add tolerations to network plugins and coredns

galal-hussein:
https://github.com/rancher/rke2/issues/1061

### DIFF
--- a/packages/rke2-sriov/charts/templates/deployment.yaml
+++ b/packages/rke2-sriov/charts/templates/deployment.yaml
@@ -27,11 +27,12 @@ spec:
       labels:
         name: sriov-network-operator
     spec:
-      nodeSelector:
-        node-role.kubernetes.io/master: "true"
       tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
+        operator: Exists
+      - effect: NoSchedule
+        key: node-role.kubernetes.io/control-plane
         operator: Exists
       serviceAccountName: sriov-network-operator
       containers:
--- a/packages/rke2-sriov/charts/templates/deployment.yaml
+++ b/packages/rke2-sriov/charts/templates/deployment.yaml
@@ -27,6 +27,8 @@ spec:
       labels:
         name: sriov-network-operator
     spec:
+      nodeSelector:
+        node-role.kubernetes.io/control-plane: "true"
       tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
--- a/packages/cilium/generated-changes/patch/values.yaml.patch
+++ b/packages/cilium/generated-changes/patch/values.yaml.patch
@@ -1,0 +1,11 @@
+--- charts-original/values.yaml
++++ charts/values.yaml
+@@ -1320,6 +1320,8 @@
+   - effect: NoSchedule
+     key: node-role.kubernetes.io/master
+   - effect: NoSchedule
++    key: node-role.kubernetes.io/control-plane
++  - effect: NoSchedule
+     key: node.cloudprovider.kubernetes.io/uninitialized
+     value: "true"
+   - key: CriticalAddonsOnly
--- a/packages/cilium/package.yaml
+++ b/packages/cilium/package.yaml
@@ -1,5 +1,5 @@
 url: https://helm.cilium.io/cilium-1.9.6.tgz
-packageVersion: 05
+packageVersion: 06
 releaseCandidateVersion: 00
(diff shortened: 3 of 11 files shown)


packages/rke2-calico/generated-changes/patch/templates/tigera-operator/02-tigera-operator.yaml.patch
```diff
@@ -1,6 +1,18 @@
 --- charts-original/templates/tigera-operator/02-tigera-operator.yaml
 +++ charts/templates/tigera-operator/02-tigera-operator.yaml
-@@ -29,7 +29,7 @@
+@@ -23,13 +23,19 @@
+           operator: Exists
+         - effect: NoSchedule
+           operator: Exists
++        - key: "node-role.kubernetes.io/control-plane"
```
brandond, line 7: This appears to already tolerate all NoExecute/NoSchedule taints, regardless of key?

packages/rke2-canal/charts/templates/daemonset.yaml
```diff
@@ -41,6 +41,12 @@ spec:
           operator: Exists
         - effect: NoExecute
           operator: Exists
+        - key: "node-role.kubernetes.io/control-plane"
```
brandond, line 44: This appears to already tolerate all NoExecute/NoSchedule taints, regardless of key?

packages/rke2-kube-proxy-1.19/charts/templates/daemonset.yaml
```diff
@@ -56,6 +56,12 @@ spec:
       - key: CriticalAddonsOnly
         operator: Exists
       - operator: Exists
```
brandond, line 58: Already appears to tolerate everything (Exists with no Key/Effect)

packages/rke2-kube-proxy-1.20/charts/templates/daemonset.yaml
```diff
@@ -56,6 +56,12 @@ spec:
       - key: CriticalAddonsOnly
         operator: Exists
       - operator: Exists
```
brandond, line 58: Already appears to tolerate everything (Exists with no Key/Effect)

packages/rke2-kube-proxy-1.21/charts/templates/daemonset.yaml
```diff
@@ -56,6 +56,12 @@ spec:
       - key: CriticalAddonsOnly
         operator: Exists
       - operator: Exists
```
brandond, line 58: Already appears to tolerate everything (Exists with no Key/Effect)

brandond (changes requested):

packages/rke2-calico/generated-changes/patch/templates/tigera-operator/02-tigera-operator.yaml.patch
galal-hussein: well yeah I thought of that too, take a look at this https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.5/rke/templates/coredns.go#L123-L128 this is for coredns in rke1 and this still doesnt tolerate the role taint, according to the docs:

```
There are two special cases:

An empty key with operator Exists matches all keys, values and effects which means this will tolerate everything.

An empty effect matches all effects with key key1.
```
But empty key with effect doesnt seem to cut it

packages/rke2-sriov/charts/templates/deployment.yaml
```diff
@@ -27,12 +27,13 @@ spec:
       labels:
         name: sriov-network-operator
     spec:
-      nodeSelector:
-        node-role.kubernetes.io/master: "true"
```
brandond, line 31: It looks like the upstream chart intended for this deployment to only run on control-plane nodes, do we need to maintain that?

galal-hussein: yes thats why I removed the nodeselector since the master role label is deprecated and added another toleration for control-plane, so that it can be scheduled on either labels

brandond: Right, you added a toleration so it **can** be scheduled there, but you didn't add a new nodeSelector for the control-plane role, so now it'll just run anywhere - which I don't think is correct.

brandond (approved):

erikwilson (approved):

briandowns (approved):

manuelbuil (approved):

PR #101, opened 2021-06-07
remove nameOverride from the ingress-nginx chart

Oats87:
https://github.com/rancher/rke2/issues/1093

Signed-off-by: Chris Kim <oats87g@gmail.com>


Oats87:
This is reverting the nameOverride that was originally added here: https://github.com/rancher/rke2-charts/pull/79

brandond (approved):

PR #105, opened 2021-06-17
rke2-kube-proxy: v1.19.12

dweomer:
Signed-off-by: Jacob Blain Christen <jacob@rancher.com>

### DIFF
--- a/packages/rke2-kube-proxy-1.19/package.yaml
+++ b/packages/rke2-kube-proxy-1.19/package.yaml
@@ -1,4 +1,4 @@
 url: local
-packageVersion: 06
+packageVersion: 01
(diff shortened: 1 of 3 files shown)


packages/rke2-kube-proxy-1.19/package.yaml
```diff
@@ -1,4 +1,4 @@
 url: local
-packageVersion: 06
+packageVersion: 01
```
briandowns, line 2: Should this be `07`?

briandowns: Convo via Slack. Thanks @dweomer !

briandowns (approved):

PR #111, opened 2021-06-24
Update rke2 calico config for Windows support

rosskirkpat:
Adding configuration into the Calico rke2 chart that's required for Windows on 2.6/rke2

### DIFF
--- a/packages/rke2-calico/generated-changes/patch/values.yaml.patch
+++ b/packages/rke2-calico/generated-changes/patch/values.yaml.patch
@@ -1,6 +1,6 @@
 --- charts-original/values.yaml
 +++ charts/values.yaml
-@@ -1,8 +1,24 @@
+@@ -1,8 +1,35 @@
  imagePullSecrets: {}
  
  installation:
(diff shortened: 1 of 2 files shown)


rosskirkpat:
@brandond @briandowns @dweomer Could I get a review on this? Can't seem to request it via the GH UI

phillipsj (approved):

dweomer:
So BGP disabled with VXLAN on by default. I don't think that breaks any assumptions for existing installations but I will defer to @briandowns and @manuelbuil

briandowns:
> So BGP disabled with VXLAN on by default. I don't think that breaks any assumptions for existing installations but I will defer to @briandowns and @manuelbuil

Looks like that's inline with current expectations.

manuelbuil:
> > So BGP disabled with VXLAN on by default. I don't think that breaks any assumptions for existing installations but I will defer to @briandowns and @manuelbuil
> 
> Looks like that's inline with current expectations.

Correct. That's the config that was agreed in the design meeting

packages/rke2-calico/generated-changes/patch/values.yaml.patch
```diff
@@ -17,15 +17,26 @@
...
++felixconfigurations:
++  spec:
++    allowVXLANPacketsFromWorkloads: true
++    vxlanEnabled: true
++    ipipEnabled: false
++    vxlanVNI: 4096
++
++ipamconfigs:
++  spec:
++    strictAffinity: true
```
manuelbuil, line 35: Unfortunately, I don't think you can pass these parameters to the operator because it does not understand them. IIRC, the current operator of Tigera has no way to fill a felixconfiguration or an ipamconfig ==> https://docs.projectcalico.org/reference/installation/api#operator.tigera.io/v1.Provider

PR #113, opened 2021-07-08
Hardened ingress-nginx 3.34.0

erikwilson:
For https://github.com/rancher/rke2/issues/659

### DIFF
--- a/packages/rke2-ingress-nginx/generated-changes/patch/values.yaml.patch
+++ b/packages/rke2-ingress-nginx/generated-changes/patch/values.yaml.patch
@@ -1,18 +1,31 @@
 --- charts-original/values.yaml
 +++ charts/values.yaml
-@@ -10,9 +10,8 @@
+@@ -10,13 +10,11 @@
  controller:
    name: controller
    image:
--    repository: k8s.gcr.io/ingress-nginx/controller
--    tag: "v0.46.0"
--    digest: sha256:52f0058bed0a17ab0fb35628ba97e8d52b5d32299fbc03cc0f6c7b9ff036b61a
+-    registry: k8s.gcr.io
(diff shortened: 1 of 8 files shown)


packages/rke2-ingress-nginx/generated-changes/patch/values.yaml.patch
```diff
@@ -1,18 +1,31 @@
...
      pullPolicy: IfNotPresent
      # www-data -> uid 101
      runAsUser: 101
-@@ -45,7 +44,7 @@
+@@ -26,7 +24,7 @@
+   existingPsp: ""
+ 
+   # Configures the controller container name
+-  containerName: controller
++  containerName: rke2-ingress-nginx-controller
```
brandond, line 24: this is just the container within the pod; curious why this needed to be changed.

brandond (approved):

brandond (approved):
LGTM when CI passes

packages/rke2-ingress-nginx/generated-changes/patch/values.yaml.patch
erikwilson: it is, they changed to `- name: {{ .Values.controller.containerName }}` in ./packages/rke2-ingress-nginx/charts/templates/controller-{daemonset,deployment}.yaml

briandowns (approved):

packages/rke2-ingress-nginx/generated-changes/patch/values.yaml.patch
erikwilson: I believe it is a leftover from upgrading older versions and the RKE2 CI checks for containers with specific names

PR #117, opened 2021-07-19
Update kube proxy versions

galal-hussein:


dereknola (approved):

manuelbuil:
Don't we need to increase packageVersion everytime we do changes? I always do that :)

manuelbuil (approved):

briandowns (approved):

PR #118, opened 2021-07-20
Add nodelocal manifests

manuelbuil:
Linked issue: https://github.com/rancher/rke2/issues/1024

### DIFF
--- a/packages/rke2-coredns/generated-changes/overlay/templates/configmap-nodelocal.yaml
+++ b/packages/rke2-coredns/generated-changes/overlay/templates/configmap-nodelocal.yaml
@@ -1,0 +1,60 @@
+{{- if .Values.nodelocal.enabled }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: node-local-dns
+  namespace: kube-system
+  labels:
+    addonmanager.kubernetes.io/mode: Reconcile
+data:
+  Corefile: |
+    {{ coalesce .Values.global.clusterDomain "cluster.local" }}:53 {
+        errors
+        cache {
+                success 9984 30
+                denial 9984 5
+        }
+        reload
+        loop
+        bind {{ .Values.nodelocal.ip_address }} {{ template "nodelocalUpstreamDNSServerIP" . }}
+        forward . {{ ternary (include "clusterDNSServerIP" .) "__PILLAR__CLUSTER__DNS__" .Values.nodelocal.ipvs  }} {
+                force_tcp
+        }
+        prometheus :9253
+        health {{ .Values.nodelocal.ip_address }}:8080
+        }
+    in-addr.arpa:53 {
+        errors
+        cache 30
+        reload
+        loop
+        bind {{ .Values.nodelocal.ip_address }} {{ template "nodelocalUpstreamDNSServerIP" . }}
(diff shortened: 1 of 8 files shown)


packages/rke2-coredns/generated-changes/overlay/templates/configmap-nodelocal.yaml
```diff
@@ -0,0 +1,60 @@
...
+    {{ coalesce .Values.global.clusterDomain "cluster.local" }}:53 {
+        errors
+        cache {
+                success 9984 30
+                denial 9984 5
+        }
+        reload
+        loop
+        bind {{ .Values.nodelocal.ip_address }} {{ template "nodelocalUpstreamDNSServerIP" . }}
+        forward . __PILLAR__CLUSTER__DNS__ {
```
brandond, line 21: As per the docs, it looks like these are supposed to be replaced with cluster-specific values before installing - so we would need to do that with Helm templating.

https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/#:~:text=if%20kube-proxy%20is%20running%20in%20iptables%20mode

brandond (changes requested):

packages/rke2-coredns/generated-changes/overlay/templates/configmap-nodelocal.yaml
manuelbuil: Yes and no. There are some that need to be replaced, e.g. __PILLAR__LOCAL__DNS__, which is done by sed in the link. However, __PILLAR__CLUSTER__DNS__ gets updated by the pod itself. That's the ip address that the service kube-dns-upstream is taking and can't be known in advanced. You can have a look at rke for comparison (they are missing the ipvs support though) => https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6/rke/templates/nodelocal.go#L45-L94

brandond: Do the pods themselves run that sed command then? I'm confused why the docs call out running that command.

manuelbuil: If you look at the first sed command in the docs, it replaces `__PILLAR__LOCAL__DNS__`, `__PILLAR__DNS__DOMAIN__` and `__PILLAR__DNS__SERVER__`. But it does not touch `__PILLAR__CLUSTER__DNS__`, so that one stays like that in the configmap. When the pod consumes it, the pod must replace it with sed because in the logs I can see how it changes to the ip of the new service:
```
    bind 169.254.20.10 
    forward . 10.43.8.123 {
            force_tcp
    }
```
Nevertheless, by carefully looking at the ipvs sed command, I have just realized that `__PILLAR__CLUSTER__DNS__` gets replaced there. So you are right, we must use templating for that too. The value will be the string: `__PILLAR__CLUSTER__DNS__` if ipvs=false and the clusterip of dns service if ipvs=true

manuelbuil: To complete the explanation, `__PILLAR__UPSTREAM__SERVERS__` does not get replaced in ipvs or non-ipvs mode. The pod does the replacement when consuming it

erikwilson (approved):

brandond (approved):

PR #119, opened 2021-07-20
Use kubernetes image instead of kube-proxy image

galal-hussein:


packages/rke2-kube-proxy-1.21/charts/values.yaml
```diff
@@ -2,8 +2,8 @@
 
 # image for kubeproxy
 image:
-  repository: rancher/hardened-kube-proxy
-  tag: v1.21.3-build20210716
+  repository: rancher/hardened-kubernetes
+  tag: v1.21.3-rke2r5-build20210716
```
briandowns, line 6: Should this be `r2`?

galal-hussein: yeah we went through different iterations until we got the right kubernetes image here you go https://github.com/rancher/image-build-kubernetes/releases

brandond (approved):

briandowns (approved):

PR #124, opened 2021-07-30
Update rke2 calico config for Windows support

manuelbuil:
Gives an alternative to https://github.com/rancher/rke2-charts/pull/111 which will not work due to the Tigera helm chart created for Calico.

Changes:
- All crds (esp. ipamconfigs) are moved into the calico-crd chart
- Required config of ipamconfigs for windows to work gets deployed in the calico chart. My initial tests show that this config does not break linux
- As strictAffinity=true and nodes cannot borrow ips, we increase the amount of usable ips per node from 62 (/24) to 126 (/25) to honor the default k8s config of 110 pods/node.


manuelbuil:
Tested in linux:

- Ping internode works
- ping intranode works
- network policy works

rosskirkpat (approved):
If I understand correctly, we would need to monitor upstream changes to the CRDs and merge those in as needed? We would effectively be maintaining a custom Calico chart with this PR. 

I believe this is our only option at this time and have no issue approving this as these changes are required for 2.6.0 functionality because the custom configuration needs deployed before any pods are created.

brandond (approved):

manuelbuil:
@luthermonson @rosskirkpat Is `allowVXLANPacketsFromWorkloads: true` also a requirement for windows support? Is there a link in the Calico docs that we could write in the comments to point to this required values? That will help in the future to maintain this chart

packages/rke2-calico/generated-changes/patch/values.yaml.patch
```diff
@@ -41,3 +41,10 @@
 +
 +global:
 +  systemDefaultRegistry: ""
++
++ipamConfig:
++  strictAffinity: true
++  autoAllocateBlocks: true
```
luthermonson, line 47: this is actually not needed, it defaults to true any way so can be dropped

packages/rke2-calico/generated-changes/patch/values.yaml.patch
```diff
@@ -41,3 +41,10 @@
...
++
++felixConfiguration:
++  allowVXLANPacketsFromWorkloads: true
```
luthermonson, line 50: this defaults to false and is not required to for strict affinity

luthermonson:
@manuelbuil didnt we talk about doing a /24 for block size?

luthermonson:
docs: https://docs.projectcalico.org/getting-started/windows-calico/kubernetes/standard#install-calico-on-linux-control-and-worker-nodes

packages/rke2-calico/generated-changes/patch/values.yaml.patch
manuelbuil: Ok. Then I suggest leaving the default

packages/rke2-calico/generated-changes/patch/values.yaml.patch
manuelbuil: We need that parameter when defining an ipamConfig:
```
            required:
            - autoAllocateBlocks
            - strictAffinity
```

manuelbuil:
> @manuelbuil didnt we talk about doing a /24 for block size?

You are right. It was because according to k8s docs 110 is the default pods per node, right? So we would need /25 or lower

brandond (approved):

galal-hussein (approved):

luthermonson:
https://github.com/rancher/rke2/issues/1289

PR #126, opened 2021-08-03
Update dns cache image

manuelbuil:
The old image did not have iptables binaries in it and we need them


brandond (approved):

manuelbuil:
I think this applies for only one review

PR #138, opened 2021-08-27
Add support for dual-stack in Calico chart

manuelbuil:
Linked issue: https://github.com/rancher/rke2/issues/1717

The logic should do the following if there is at least one ipv6 cidr:
- Enable BGP
- Remove the VXLAN encapsulation from ippools
- Create a new ippool dictionary with the ipv6 cidr

### DIFF
--- a/packages/rke2-calico/generated-changes/patch/templates/crs/custom-resources.yaml.patch
+++ b/packages/rke2-calico/generated-changes/patch/templates/crs/custom-resources.yaml.patch
@@ -1,6 +1,6 @@
 --- charts-original/templates/crs/custom-resources.yaml
 +++ charts/templates/crs/custom-resources.yaml
-@@ -6,6 +6,13 @@
+@@ -6,6 +6,27 @@
  {{ $secrets = append $secrets $item }}
  {{ end }}
  {{ $_ := set $installSpec "imagePullSecrets" $secrets }}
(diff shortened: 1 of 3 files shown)


vadorovsky (approved):

cjellick:
We need to make sure this is backwards compatible.

brandond:
@cjellick wouldn't the appropriate way to set these be by setting the RKE2 --cluster-cidr and --service-cidr flags?

cjellick:
I had to fire and forget that backwards comapt comment while i was in a meeting. @cloudnautique has a user headed down the calico + dual stack path nowish. If they get it working on a version that is before the version this lands it, we don't want it to break.

Do you all need more details in order to understand how they're deploying or anything like that?

brandond:
If they've created their own IP Pools I don't think this would interfere with that, but I'm not sure how Calico would handle duplicate or conflicting configurations. Knowing how they're doing it now and how far they got with it would probably be good.

manuelbuil:
> If they've created their own IP Pools I don't think this would interfere with that, but I'm not sure how Calico would handle duplicate or conflicting pools. Knowing how they're doing it now and how far they got with it would probably be good.

Agree to this. Let me do some quick tests to verify what happens

manuelbuil:
These are the 3 upgrade use cases I could think of. I tested them:

1 - User deploys rke2 with `cluster-cidr: 10.42.0.0/16,2001:cafe:42:0::/56` and the old calico chart (without this PR). Result ==> Calico does not work and calico-node stays in `CrashLoopBackOff`

2 - User deploys rke2 with `cluster-cidr: 10.42.0.0/16` and then, in an upgrade, tries to move to dual-stack config (ipv6 cidr, bgp...). Result ==> Things don't work because the operator does not change the created resources, i.e. vxlan config stays in ippools, felixconfiguration.... According to Tigera, manual intervention is needed from Calico perspective

3 - User deploys rke2 with `cluster-cidr: 10.42.0.0/16,2001:cafe:42:0::/56` and without a cni plugin. Then installs Calico from upstream with the correct config to do dual-stack from the beginning. Then, tries to upgrade using rke2 helmController. Result ==> Things keep working. The operator does not touch the already created ippools

Conclusion: When doing upgrades, Calico operator leaves the ippools as they are, even if the new config says otherwise (e.g. change encapsulation or add new ippool). If changes in the Calico internals are needed, human intervention is required to edit the appropriate calico resources. Waiting for confirmation from Tigera to validate this conclusion

manuelbuil:
Note that according to @rosskirkpat dual-stack is not in a functional state for k8s+windows. Moreover, the windows team has not tested BGP yet

brandond (approved):

PR #140, opened 2021-09-01
Add harvester-cloud-provider and harvester-csi-driver

yaocw2020:
issue: https://github.com/rancher/rke2/issues/1680

### DIFF
--- a/packages/harvester-cloud-provider/package.yaml
+++ b/packages/harvester-cloud-provider/package.yaml
@@ -1,0 +1,3 @@
+url: https://github.com/harvester/charts/releases/download/harvester-cloud-provider-0.1.2/harvester-cloud-provider-0.1.2.tgz
+packageVersion: 00
+releaseCandidateVersion: 00
--- a/packages/harvester-csi-driver/package.yaml
+++ b/packages/harvester-csi-driver/package.yaml
@@ -1,0 +1,3 @@
+url: https://github.com/harvester/charts/releases/download/harvester-csi-driver-0.1.3/harvester-csi-driver-0.1.3.tgz
+packageVersion: 00
+releaseCandidateVersion: 00


dweomer (changes requested):
Please see https://github.com/rancher/rke2-charts#basic-commands for instructions on the steps to perform to submit a PR. Specifically, it looks as if `make clean` was not run as there are `assets/*` and `charts/*` and `index.yaml` in this PR. The automation takes care of this for us, please do not include it in your PR.

yaocw2020:
> Please see https://github.com/rancher/rke2-charts#basic-commands for instructions on the steps to perform to submit a PR. Specifically, it looks as if `make clean` was not run as there are `assets/*` and `charts/*` and `index.yaml` in this PR. The automation takes care of this for us, please do not include it in your PR.

Oh, it is so convenient. And what's the version form generated by automation?

dweomer (approved):

brandond:
We inject a number of cluster configuration values into charts; the most important of which is the systemDefaultRegistry. If set this should be used as a registry prefix for all the images used by the chart.

The list of variables is listed here:
* https://github.com/rancher/rke2/blob/82ca956e2d43e7e10223375c7e9614ac3acf6d0c/pkg/bootstrap/bootstrap.go#L268-L274

For an example of other charts using this you can look at
* https://github.com/rancher/rke2-charts/blob/de5dd4c9f4ded7c08b4417922f691139d2417e9f/packages/rke2-canal/charts/templates/_helpers.tpl#L1-L7
* https://github.com/rancher/rke2-charts/blob/de5dd4c9f4ded7c08b4417922f691139d2417e9f/packages/rke2-canal/charts/templates/daemonset.yaml#L53

brandond (changes requested):
needs system-default-registry support as describe above.

yaocw2020:
> needs system-default-registry support as describe above.

Done! @brandond

guangbochen:
the chart is updated and it was addressed via https://github.com/harvester/charts/pull/22, thanks.

brandond (approved):
lgtm!

PR #148, opened 2021-09-24
Explicitly disable encapsulation in dual-stack env for Calico

manuelbuil:
When not specified, the encapsulation of ipv4 traffic becomes IPIP.

That encapsulation is the least recommended one by Calico ==> https://docs.projectcalico.org/networking/determine-best-networking

Linked issue: https://github.com/rancher/rke2/issues/1876


dereknola (approved):

vadorovsky (changes requested):
Typo in commit message:

Explicit**e**ly -> Explicitly

From my understanding - you are disabling encapsulation only for IPv6?

manuelbuil:
> Typo in commit message:
> 
> Explicit**e**ly -> Explicitly
> 
> From my understanding - you are disabling encapsulation only for IPv6?

The other way around. ipv6 does not have any encapsulation, I'm disabling it for ipv4. For ipv6:

+{{ $myIP6Dict := dict "natOutgoing" "Enabled" "cidr" .Values.global.clusterCIDRv6 }}

I could add another field "Encapsulation" "False" but it is not needed

vadorovsky (approved):
Ah, right. LGTM now!

PR #156, opened 2021-10-06
Upgrade nginx controller to v1.0.2

galal-hussein:
Since the helm chart for v1.0.2 was also updated upstream, I had to delete the patches and prepare as if its a new chart, then added our patches to the chart

### DIFF
--- a/packages/rke2-ingress-nginx/generated-changes/patch/templates/_helpers.tpl.patch
+++ b/packages/rke2-ingress-nginx/generated-changes/patch/templates/_helpers.tpl.patch
@@ -25,3 +25,4 @@
 +{{- define "repository_or_registry_and_image" -}}
 +{{- if .repository -}}{{ .repository }}{{ else }}{{ .registry }}/{{ .image }}{{- end -}}:{{ .tag }}{{- if (.digest) -}} @{{.digest}} {{- end -}}
 +{{- end -}}
+\ No newline at end of file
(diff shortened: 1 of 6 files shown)


packages/rke2-ingress-nginx/generated-changes/patch/templates/_helpers.tpl.patch
```diff
@@ -25,3 +25,4 @@
 +{{- define "repository_or_registry_and_image" -}}
 +{{- if .repository -}}{{ .repository }}{{ else }}{{ .registry }}/{{ .image }}{{- end -}}:{{ .tag }}{{- if (.digest) -}} @{{.digest}} {{- end -}}
 +{{- end -}}
+\ No newline at end of file
```
briandowns, line 28: Should a new line have been added here?

galal-hussein: removing it now

packages/rke2-ingress-nginx/generated-changes/patch/templates/controller-deployment.yaml.patch
```diff
@@ -1,6 +1,15 @@
 --- charts-original/templates/controller-deployment.yaml
 +++ charts/templates/controller-deployment.yaml
-@@ -68,9 +68,7 @@
+@@ -23,7 +23,7 @@
+   replicas: {{ .Values.controller.replicaCount }}
+   {{- end }}
+   revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
+-  {{- if .Values.controller.updateStrategy }}
++  {{- if .Values.controller.updateStrfategy }}
```
brandond, line 8:
```suggestion
+  {{- if .Values.controller.updateStrategy }}
```

PR #184, opened 2021-11-01
Bump harvester cloud provider and csi driver

yaocw2020:


guangbochen (approved):
LGTM

brandond (changes requested):
Please modify the existing packages in-place rather than creating new versioned ones parallel to the existing unversioned packages.

yaocw2020:
> Please modify the existing packages in-place rather than creating new versioned ones parallel to the existing unversioned packages.

Done

brandond (approved):

PR #185, opened 2021-11-08
Bump coredns to 1.16.4

olljanat:
Provider coredns extra config support from https://github.com/coredns/helm/pull/44 and solves my feature request on https://github.com/rancher/rke2/issues/1808

### DIFF
--- a/packages/rke2-coredns/package.yaml
+++ b/packages/rke2-coredns/package.yaml
@@ -1,4 +1,4 @@
-url: https://github.com/coredns/helm/releases/download/coredns-1.16.3/coredns-1.16.3.tgz
-packageVersion: 02
+url: https://github.com/coredns/helm/releases/download/coredns-1.16.4/coredns-1.16.4.tgz
+packageVersion: 00
 # This repository does not use releaseCandidateVersions, so you can leave this as 00.
 releaseCandidateVersion: 00


packages/rke2-coredns/package.yaml
```diff
@@ -1,4 +1,4 @@
-url: https://github.com/coredns/helm/releases/download/coredns-1.16.3/coredns-1.16.3.tgz
-packageVersion: 02
+url: https://github.com/coredns/helm/releases/download/coredns-1.16.4/coredns-1.16.4.tgz
+packageVersion: 03
```
brandond, line 2: You can drop this back down to 0 when bumping the upstream chart version; the package version only needs to be incremented if changing the local customizations without bumping the upstream chart version.
```suggestion
packageVersion: 00
```

brandond (changes requested):

packages/rke2-coredns/package.yaml
olljanat: Ok. Fixed

brandond (approved):

olljanat:
ping @manuelbuil I see that you have you have reviewed coredns related changes earlier.

manuelbuil (approved):
LGTM! After this is merged, we need to update rke2 to use it

PR #196, opened 2021-12-08
bump kube-proxy

raulcabello:


brandond:
Closing this; will get done as part of the upcoming release cycle.

PR #199, opened 2021-12-13
Added autogenerated PSK field and change type to Secret

dgiebert:
As specified in the [flannel docs](https://github.com/flannel-io/flannel/blob/master/Documentation/backends.md#ipsec) the ipsec variant needs to have a pre-shared key.

Also related to https://github.com/rancher/image-build-flannel/issues/7

### DIFF
--- a/packages/rke2-canal/charts/templates/config.yaml
+++ b/packages/rke2-canal/charts/templates/config.yaml
@@ -8,7 +8,7 @@ metadata:
   namespace: kube-system
 stringData:
 {{- $config := (lookup "v1" "Secret" .Release.Namespace "{{ .Release.Name }}-config)") | default dict }}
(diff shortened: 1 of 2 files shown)


manuelbuil:
Sorry, I totally missed this PR! 

Note that we are basically consuming upstream canal with really minor differences. This PR will increase those differences and thus I wonder if you could first contribute these changes to the upstream project and then we can incorporate them here, otherwise maintenance might become difficult

dgiebert:
Can surely do, could you point me to the correct repo?

manuelbuil:
> Can surely do, could you point me to the correct repo?

Yes! https://github.com/projectcalico/calico/tree/master/calico/_includes/charts/calico. Thanks a lot

dereknola:
Closing as stale

PR #209, opened 2022-01-14
Provide dual-stack support to canal

manuelbuil:
Update the calico image to `v3.20.3-build20220114`, which supports dual-stack.

Change the config.yaml so that the ipv6 config only applies when clustercidr contains ipv6

Issue: https://github.com/rancher/rke2/issues/2342


manuelbuil:
Tested and dualStack seems to work

rosskirkpat (approved):
LGTM!

rbrtbnfgl (approved):

PR #234, opened 2022-03-21
Updating CPI to have an image override to use 1.22 on 1.23.

phillipsj:
Since the chart uses a template helper to determine the CPI image version to use, we need to specify an image override by patching the chart for 1.23.

### DIFF
--- a/packages/rancher-vsphere-cpi/generated-changes/patch/values.yaml.patch
+++ b/packages/rancher-vsphere-cpi/generated-changes/patch/values.yaml.patch
@@ -1,0 +1,12 @@
+--- charts-original/values.yaml
++++ charts/values.yaml
+@@ -11,6 +11,9 @@
+ 
+ cloudControllerManager:
+   image:
++    "1.23":
++      repository: rancher/mirrored-cloud-provider-vsphere-cpi-release-manager
++      tag: v1.22.5
+     "1.22":
+       repository: rancher/mirrored-cloud-provider-vsphere-cpi-release-manager
+       tag: v1.22.5
--- a/packages/rancher-vsphere-cpi/package.yaml
+++ b/packages/rancher-vsphere-cpi/package.yaml
@@ -1,4 +1,4 @@
 url: https://github.com/rancher/vsphere-charts.git
 subdirectory: charts/rancher-vsphere-cpi
 commit: 81da645ddf86370fc5d1e35dc4711a45b3cea929
-version: 100.2.0
+packageVersion: 02


phillipsj:
@rosskirkpat @briandowns @luthermonson @sirredbeard @brandond @PennyScissors

briandowns (approved):

packages/rancher-vsphere-cpi/generated-changes/patch/values.yaml.patch
```diff
@@ -0,0 +1,12 @@
+--- charts-original/values.yaml
++++ charts/values.yaml
+@@ -11,6 +11,9 @@
+ 
+ cloudControllerManager:
+   image:
++    "override":
```
brandond, line 7: This looks like it'll override the image for all Kubernetes versions. Would it make more sense to just define this as the image for 1.23 clusters and let older versions still use whatever tag they would normally use?

phillipsj: I am going to defer that to @PennyScissors if that is what we should consider as an approach.

packages/rancher-vsphere-cpi/package.yaml
```diff
@@ -1,4 +1,4 @@
 url: https://github.com/rancher/vsphere-charts.git
 subdirectory: charts/rancher-vsphere-cpi
 commit: 81da645ddf86370fc5d1e35dc4711a45b3cea929
-version: 100.2.0
+version: 100.2.1
```
brandond, line 4: For the CI failure - I copied the package.yaml from rancher charts, but it looks like the version is actually ignored, and we want to do this instead:
```suggestion
packageVersion: 01
```

phillipsj: Ah the old scripts

packages/rancher-vsphere-cpi/generated-changes/patch/values.yaml.patch
PennyScissors: That's correct, this will override for all Kubernetes versions. If you want 1.23 specifically to use a 1.22 image, you could try changing "override" for "1.23", and leave the repo and tag with the 1.22 images as is.

packages/rancher-vsphere-cpi/package.yaml
PennyScissors: AFAIK packageVersion is a Rancher 2.5 thing, but my understanding is that rke2-charts doesn't have the latest charts build scripts so it may be old enough that it still uses packageVersion. This needs to be checked.

packages/rancher-vsphere-cpi/generated-changes/patch/values.yaml.patch
phillipsj: @briandowns @luthermonson @rosskirkpat any opinions?

briandowns: Let's go with specifying this as the version for 1.23 clusters only.

phillipsj: Ok, I will get it changed and pushed

briandowns (approved):

luthermonson (approved):

brandond (approved):

PR #246, opened 2022-04-27
Updated canal chart for ipv6 only config

rbrtbnfgl:
Updated Canal chart to support IPv6 only setup


manuelbuil (approved):
config.yaml looks uglier but I don't see how to make it better

brandond (approved):

PR #253, opened 2022-05-09
Updated calico chart to support IPv6 only setup

rbrtbnfgl:
Signed-off-by: Roberto Bonafiglia <roberto.bonafiglia@suse.com>

Updated calico chart to disable IPv4 in case of an IPv6 only setup.

### DIFF
--- a/packages/rke2-calico/generated-changes/patch/templates/crs/custom-resources.yaml.patch
+++ b/packages/rke2-calico/generated-changes/patch/templates/crs/custom-resources.yaml.patch
@@ -1,6 +1,6 @@
 --- charts-original/templates/crs/custom-resources.yaml
 +++ charts/templates/crs/custom-resources.yaml
-@@ -6,6 +6,27 @@
+@@ -6,6 +6,30 @@
  {{ $secrets = append $secrets $item }}
  {{ end }}
  {{ $_ := set $installSpec "imagePullSecrets" $secrets }}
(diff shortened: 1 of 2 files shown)


packages/rke2-calico/generated-changes/patch/templates/crs/custom-resources.yaml.patch
```diff
@@ -16,11 +16,14 @@
...
++{{ $finalIpPoolList := list $myIP6Dict }}
++{{ if not (empty .Values.global.clusterCIDRv4) }}
 +{{ $allIpPools := get .Values.installation.calicoNetwork "ipPools" }}
 +{{ range $allIpPools }}
 +{{ $_ := set . "encapsulation" "None" }}
 +{{ end }}
 +{{ $finalIpPoolList := append $allIpPools $myIP6Dict }}
++{ end }}
 +{{ $calicoNetwork := get .Values.installation "calicoNetwork" }}
 +{{ $_ := set $calicoNetwork "ipPools" $finalIpPoolList }}
```
manuelbuil, line 28: If `.Values.global.clusterCIDRv6` is empty, finalIpPoolList does not exist, right? Doesn't helm complain about it?

rbrtbnfgl: It will not go on that if

manuelbuil: you are right. I'm stupid

manuelbuil (approved):

brandond (approved):

PR #268, opened 2022-05-24
Add CoreDNS tolerations for cloud provider

dkeightley:
When creating a cluster with an external cloud provider controller enabled (AWS CCM for example), initially the nodes will have a `node.cloudprovider.kubernetes.io/uninitialized` taint.

This prevents CoreDNS pods from starting, and in turn the CCM is unable to resolve and initialise the node.

Adding the toleration allows CoreDNS to tolerate this condition.

Note: example here required the HelmChartConfig to get a working cluster with the AWS CCM, can be used for QA by removing the HelmChartConfig.
  * https://gist.github.com/dkeightley/26607d6739429a174675a81cd6fe65d6

### DIFF
--- a/packages/rke2-coredns/generated-changes/patch/templates/_helpers.tpl.patch
+++ b/packages/rke2-coredns/generated-changes/patch/templates/_helpers.tpl.patch
@@ -1,6 +1,6 @@
 --- charts-original/templates/_helpers.tpl
 +++ charts/templates/_helpers.tpl
-@@ -137,6 +137,7 @@
+@@ -150,6 +150,7 @@
      {{- end -}}
  {{- end -}}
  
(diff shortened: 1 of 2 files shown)


brandond:
I'm not sure that allowing coredns to run before the nodes are initialized is a good idea. My recommendation would probably be to have the CCM DaemonSet pods run with hostnework so that it's not reliant on the CNI, cluster dns, etc to all be up. There are all kinds of nasty chicken-and-egg problems if you try to make that work.

It looks like there's even a chart value for this in the AWS CCM chart:

https://github.com/kubernetes/cloud-provider-aws/blob/master/charts/aws-cloud-controller-manager/templates/daemonset.yaml#L26-L28

dkeightley:
Thanks, that seems like the best approach.

I'll test and then close this out.

PR #273, opened 2022-06-13
Bump coredns and update resources

brandond:
* Bump coredns chart version and image tag
* Increase autoscaler resources so that it isn't killed for exceeding limits

Related to:
* https://github.com/rancher/rke2/issues/2175

### DIFF
--- a/packages/rke2-coredns/generated-changes/patch/Chart.yaml.patch
+++ b/packages/rke2-coredns/generated-changes/patch/Chart.yaml.patch
@@ -1,14 +1,5 @@
 --- charts-original/Chart.yaml
 +++ charts/Chart.yaml
-@@ -2,7 +2,7 @@
-   artifacthub.io/changes: |
-     - Initial helm chart changelog
- apiVersion: v2
--appVersion: 1.8.7
-+appVersion: 1.9.1
- description: CoreDNS is a DNS server that chains plugins and provides Kubernetes DNS
-   Services
- home: https://coredns.io
 @@ -14,7 +14,7 @@
  maintainers:
  - name: mrueg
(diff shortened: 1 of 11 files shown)


packages/rke2-coredns/generated-changes/patch/Chart.yaml.patch
```diff
@@ -1,14 +1,5 @@
 --- charts-original/Chart.yaml
 +++ charts/Chart.yaml
-@@ -2,7 +2,7 @@
-   artifacthub.io/changes: |
```
briandowns, line 4: Was removing this bit intentional?

brandond: Reviewing patches to patches... we're not patching it out, it was removed from the upstream Chart.yaml. Our only change remains the appVersion.

briandowns (approved):

packages/rke2-coredns/generated-changes/patch/values.yaml.patch
```diff
@@ -123,17 +123,19 @@
...
-+      memory: "20Mi"
++      cpu: "25m"
++      memory: "16Mi"
      limits:
-       cpu: "20m"
+-      cpu: "20m"
 -      memory: "10Mi"
-+      memory: "20Mi"
++      cpu: "100m"
++      memory: "64Mi"
```
pdostal, line 134: Why is the `memory` specified two times

pdostal (commented):
I had similar problem in my rke2 cluster and increasing `cpu` and `memory` fixed the issue for me. Thank you for your patch!

PR #277, opened 2022-06-24
Updated calico chart to expose felix cfg props exposed in canal

ron1:
Updated calico chart to expose felix config properties exposed in canal chart. See issue https://github.com/rancher/rke2/issues/3091


dereknola (approved):
LGTM

dereknola (changes requested):
You need to bump the chart version https://github.com/rancher/rke2-charts/blob/main-source/packages/rke2-calico/package.yaml#L2

brandond (approved):

ron1:
@dereknola do you have a moment to give this PR a final review?

dereknola (approved):

manuelbuil:
We had to remove the fields:
```
iptablesRefreshInterval:
iptablesBackend:
```
from the felix configuration in the latest release because in Calico v3.24.1, calico-node was failing due to these parameters. I suspect there is a bug in how the calico operator is passing these values because calico-node was not recognizing a valid string for iptablesBackend or a valid integer for iptablesRefreshInterval. As passing values like this is not the recommended way by Calico, they are probably never testing it

PR #283, opened 2022-08-31
Update calico version to 3.23.3 in rke2-calico package

ruslanloman:
Hi, 
There is a bug with ip_autodetection mode "k8s-internal-ip" in calico 3.23.1 projectcalico/calico#6142 and it was fixed in 3.23.2 version projectcalico/calico#6228.

### DIFF
--- a/packages/rke2-calico/generated-changes/patch/values.yaml.patch
+++ b/packages/rke2-calico/generated-changes/patch/values.yaml.patch
@@ -31,13 +31,13 @@
  tigeraOperator:
 -  image: tigera/operator
 +  image: rancher/mirrored-calico-operator
-   version: v1.27.1
+   version: v1.27.12
 -  registry: quay.io
 +  registry: docker.io
  calicoctl:
 -  image: docker.io/calico/ctl
 +  image: rancher/mirrored-calico-ctl
-   tag: v3.23.1
+   tag: v3.23.3
 +
 +global:
 +  systemDefaultRegistry: ""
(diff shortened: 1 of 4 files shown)


brandond (approved):

packages/rke2-calico/package.yaml
```diff
@@ -1,5 +1,5 @@
-url: https://github.com/projectcalico/calico/releases/download/v3.23.1/tigera-operator-v3.23.1.tgz
-packageVersion: 04
+url: https://github.com/projectcalico/calico/releases/download/v3.23.3/tigera-operator-v3.23.3.tgz
+packageVersion: 05
```
manuelbuil, line 2: When the tarball changes, we move the packaVersion back to 01

manuelbuil (commented):
Thanks for the PR. The images need to be ready here: https://github.com/rancher/image-mirror/blob/master/images-list, I can do that for you

packages/rke2-calico/generated-changes/patch/values.yaml.patch
```diff
@@ -37,7 +37,7 @@
  calicoctl:
 -  image: docker.io/calico/ctl
 +  image: rancher/mirrored-calico-ctl
-   tag: v3.23.1
+   tag: v3.23.3
```
manuelbuil, line 40: You should also move the tigera operator version to `v1.27.12`, which seems to be the latest one to install v3.23.3 ==> [v1.27.12](https://github.com/tigera/operator/releases/tag/v1.27.12)

packages/rke2-calico/package.yaml
ruslanloman: Thanks for your comment.

packages/rke2-calico/generated-changes/patch/values.yaml.patch
ruslanloman: Thanks for your comment. Fixed.

manuelbuil (approved):

PR #299, opened 2022-11-08
Added PSP for tigera-operator for CIS-1.6

rbrtbnfgl:
Added PSP for tigera-operator in case of version <1.25 for CIS-1.6
https://github.com/rancher/rke2/issues/3532


thomasferrandiz (approved):

manuelbuil:
clusterAPI psp config would still be required but that it fine as it is not part of the default deployment

manuelbuil (approved):

PR #303, opened 2022-11-17
Add volume snapshot CRDs, controller, and validation webhook

brandond:
Uses the well-maintained charts from [piraeusdatastore](https://github.com/piraeusdatastore/helm-charts/tree/main/charts/) because upstream does not want to maintain their own charts, as per https://github.com/kubernetes-csi/external-snapshotter/pull/622#issuecomment-1122649507

These charts are Apache licensed, and safe to pull in to RKE2.

This also repackages the CRDs into a separate chart, as is the pattern for other charts like rke2-calico.


dereknola (approved):

sureshgoli25:
We are looking forward for this pull request to be merged and waiting for rke2 release with this pull request.

Will this pull request targets which RKE2 releases?

brandond:
At this point I would guess that it might make it for the February releases.

briandowns (approved):

PR #320, opened 2023-01-18
Update Cilium to v1.12.5 and Hubble to v0.10.0

pjbgf:


brandond (approved):

packages/rke2-cilium/package.yaml
```diff
@@ -1,2 +1,2 @@
 url: https://helm.cilium.io/cilium-1.12.4.tgz
```
rbrtbnfgl, line 1:
```suggestion
url: https://helm.cilium.io/cilium-1.12.5.tgz
```

packages/rke2-cilium/package.yaml
```diff
@@ -1,2 +1,2 @@
...
-packageVersion: 02
+packageVersion: 03
```
rbrtbnfgl, line 2:
```suggestion
packageVersion: 01
```

packages/rke2-cilium/package.yaml
brandond, line 1: the chart version needs to be bumped to 1.12.5 as per https://github.com/rancher/rke2/pull/3802/files#r1073763146.

When doing this, please reset the packageVersion to 00.

brandond (changes requested):

rbrtbnfgl:
There are few file that you are patching but I think we don't need to patch. Probably you generated the patch from a `1.12.4` to a `1.12.5` and it generates the patch also to the modification done by Cilium.

rbrtbnfgl:
Do we need to change hubble-ui version on the chart? Usually we use the same version of the images from the source chart.

pjbgf:
That documentation comes from the Hubble repository: https://github.com/cilium/hubble#releases.

Potentially the reason why the chart hasn't been updated lately is that most folks use [cilium-cli](https://github.com/cilium/cilium-cli) for deploying Cilium and its sub-components.

rbrtbnfgl (approved):

brandond (approved):

PR #323, opened 2023-02-15
Bump ingress-nginx helm chart to 4.3.0

dereknola:
This is a transitional PR. It bumps the helm chart  we use from a old 4.1.0 (april 2022) to 4.3.0 (Oct 2022) which was the last helm chart that shipped with a controller-v1.4.X (ie our nginx-1.4.1-hardened2). It **does not** bump the actual controller itself

Some notes:
- This pulls in https://github.com/kubernetes/ingress-nginx/pull/8733 and https://github.com/kubernetes/ingress-nginx/pull/8890, which covers most of the bugs in `clusterrole.yaml`, so a smaller patch is now needed.
- The upstream helm chart now pulls from `registry.k8s.io`, they moved away from `k8s.gcr.io`. Doesn't affect us, but makes our patches look slightly different going forward.

Once this merges and is tested with a full rke2 release, we can move on to the most recent helm chart 4.5.2 and the contoller-v1.6.X (ie our nginx-1.6.4-hardened1)

### DIFF
--- a/packages/rke2-ingress-nginx/generated-changes/patch/Chart.yaml.patch
+++ b/packages/rke2-ingress-nginx/generated-changes/patch/Chart.yaml.patch
@@ -1,6 +1,6 @@
 --- charts-original/Chart.yaml
 +++ charts/Chart.yaml
-@@ -28,7 +28,7 @@
+@@ -19,7 +19,7 @@
  - name: rikatz
  - name: strongjz
  - name: tao12345666333
@@ -8,4 +8,4 @@
 +name: rke2-ingress-nginx
  sources:
  - https://github.com/kubernetes/ingress-nginx
- type: application
+ version: 4.3.0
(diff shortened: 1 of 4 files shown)


brandond (changes requested):

brandond (approved):

boris-stojnev:
Hi @dereknola , 

I was trying to install ingress-nginx v4.3.0 and wasn’t able to. After spending some time, I figured out that you changed the tag version for the rancher/mirrored-ingress-nginx-kube-webhook-certgen image to v1.5.2 that is not available from the docker hub under rancher repo. https://hub.docker.com/r/rancher/mirrored-ingress-nginx-kube-webhook-certgen/tags

I installed v4.1.0 for now which is using image tag v1.1.1. Upgrading is not possible as well. 

Is there a workaround?

Thanks.

PR #328, opened 2023-03-09
Update flannel version and values on canal

rbrtbnfgl:
Update Flannel version to `v0.21.3`


dereknola (approved):

packages/rke2-canal/charts/values.yaml
```diff
@@ -20,7 +20,32 @@ flannel:
   # Backend for kube-flannel. Backend should not be changed
   # at runtime.
   backend: "vxlan"
-  backendPort: 0 # When 0 use default port
+  # Port used by the backend 0 means default value (VXLAN: 8472, Wireguard: 51821, UDP: 8285)
+  backendPort: 0
+  # MTU to use for outgoing packets (VXLAN and Wiregurad) if not defined the MTU of the external interface is used.
```
brandond, line 25:
```suggestion
  # MTU to use for outgoing packets (VXLAN and Wireguard) if not defined the MTU of the external interface is used.
```

packages/rke2-canal/charts/values.yaml
```diff
@@ -20,7 +20,32 @@ flannel:
...
+  # VXLAN Identifier to be used. On Linux default is 1.
+  #vni: 1
+  # Enable VXLAN Group Based Policy (Default false)
+  GBP: false
+  # Enable direct routes (default is false)
+  directRouting: false
+  # MAC prefix to be used on Windows. (Defaults is 0E-2A)
+  #macPrefix: "0E-2A"
+  #
+  # Wireguard Configs:
```
brandond, line 39: is this section empty? The comment here doesn't seem to pertain to anything.

PR #343, opened 2023-06-08
Bump coredns helmchart to 1.24.0

brandond:
Does not change the image, just the chart

* For https://github.com/rancher/rke2/issues/4321


packages/rke2-coredns/generated-changes/patch/values.yaml.patch
```diff
@@ -1,17 +1,26 @@
...
 +  tag: "v1.10.1-build20230406"
    pullPolicy: IfNotPresent
    ## Optionally specify an array of imagePullSecrets.
    ## Secrets must be manually created in the namespace.
-@@ -76,10 +76,10 @@
+@@ -54,7 +54,6 @@
+ 
+ service:
+ # clusterIP: ""
+-# clusterIPs: []
```
manuelbuil, line 19: It's weird that we remove clusterIPs in the values. It's a comment though, so no harm

manuelbuil (approved):

rbrtbnfgl (approved):

packages/rke2-coredns/generated-changes/patch/values.yaml.patch
brandond: The chart update added support for ClusterIPs as a multi-value replacement for ClusterIP, but we don't support it because we set ClusterIP from the cluster dns address value.

PR #352, opened 2023-06-30
Bump ingress-nginx to 4.6.1 chart

dereknola:
Signed-off-by: Derek Nola <derek.nola@suse.com>


packages/rke2-ingress-nginx/package.yaml
```diff
@@ -1,4 +1,4 @@
-url: https://github.com/kubernetes/ingress-nginx/releases/download/helm-chart-4.5.2/ingress-nginx-4.5.2.tgz
-packageVersion: 02
+url: https://github.com/kubernetes/ingress-nginx/releases/download/helm-chart-4.6.1/ingress-nginx-4.6.1.tgz
+packageVersion: 01
```
brandond, line 2: usually we go back to 00 on bump but it doesn't really matter.

brandond (approved):

PR #362, opened 2023-08-02
Update calico and flannel on canal and added regex for the interace

rbrtbnfgl:
issue: https://github.com/rancher/rke2/issues/4515
This update also add a new value on the chart `regexIface` that permits to define a regular expression that matches the interface used by flannel on the node.

### DIFF
--- a/packages/rke2-canal/charts/templates/daemonset.yaml
+++ b/packages/rke2-canal/charts/templates/daemonset.yaml
@@ -256,6 +256,11 @@ spec:
                 configMapKeyRef:
                   name: {{ .Release.Name }}-config
                   key: canal_iface
+            - name: FLANNELD_IFACE_REGEX
+              valueFrom:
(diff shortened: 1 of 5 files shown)


packages/rke2-canal/charts/templates/daemonset.yaml
```diff
@@ -256,6 +256,11 @@ spec:
                 configMapKeyRef:
                   name: {{ .Release.Name }}-config
                   key: canal_iface
+            - name: FLANNELD_IFACE_REGEX
```
manuelbuil, line 259: Does flannel read the env variable `FLANNELD_IFACE_REGEX`?

manuelbuil (commented):
Why did you decide to use an env variable instead of using `Values.flannel.args`?

manuelbuil (approved):

PR #367, opened 2023-08-21
Change systemGlobalRegistry path

thomasferrandiz:
Change the value from `.Values.global.cattle.systemDefaultRegistry` to `.Values.global.systemDefaultRegistry` to align with the other charts.

This is required before merging the PR with validation tests in rke2 (https://github.com/rancher/rke2/pull/4615).


manuelbuil:
LGTM but I don't know how we are consuming those charts. I'd prefer someone that knows to review :)

brandond:
This shouldn't be necessary. RKE2 actually injects both values, for compatibility with Rancher - which uses a different value (don't ask me why).

https://github.com/rancher/rke2/blob/3ab96dd0016a637f48672b011822de5a35c96045/pkg/bootstrap/bootstrap.go#L312-L313

thomasferrandiz:
PR not needed once `global.cattle.systemGlobalRegistry` in injected in the test script

PR #368, opened 2023-08-22
Add prometheusMetricsEnabled option to rke2-calico

vadossam:
Add to rke2-calico FelixConfiguration template `prometheusMetricsEnabled` option which is required for enabling [metrics exporter.](https://docs.tigera.io/calico/latest/operations/monitor/monitor-component-metrics#1-configure-calico-to-enable-metrics-reporting)

PR #369, opened 2023-08-25
Added felixConfig values to be added from values.yaml file

rbrtbnfgl:
Removed the static values on `felixconfig.yaml` to dynamically get them from the `values.yaml` file to fix #368

### DIFF
--- a/packages/rke2-calico/generated-changes/patch/values.yaml.patch
+++ b/packages/rke2-calico/generated-changes/patch/values.yaml.patch
@@ -26,7 +26,7 @@
  
  certs:
    node:
-@@ -40,9 +53,30 @@
+@@ -40,9 +53,28 @@
  
  # Image and registry configuration for the tigera/operator pod.
  tigeraOperator:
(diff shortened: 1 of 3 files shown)


thomasferrandiz (approved):

brandond (approved):

packages/rke2-calico/generated-changes/patch/values.yaml.patch
```diff
@@ -56,7 +56,5 @@
 +  featureDetectOverride: "ChecksumOffloadBroken=true"
 +  healthPort: 9099
 +  defaultEndpointToHostAction: "Drop"
-+  failsafeInboundHostPorts: ""
-+  failsafeOutboundHostPorts: ""
```
manuelbuil, line 60: If you remove these two, they will not be able to be set via helm

rbrtbnfgl: Yes you can with the new logic you can specify whatever value you want through helm.

manuelbuil: ok, got it, thanks. The problem is that in Rancher UI, users will not be able to see what parameters can be changed, only the ones in values. So it is good if we keep the most common ones

manuelbuil (approved):

PR #377, opened 2023-10-10
Update for coredns, metrics-server, whereabout, multus and canal 

johnatasr:


packages/rke2-calico/package.yaml
```diff
@@ -1,5 +1,5 @@
 url: https://github.com/projectcalico/calico/releases/download/v3.26.1/tigera-operator-v3.26.1.tgz
-packageVersion: 01
+packageVersion: 02
```
thomasferrandiz, line 2: why update the `packageVersion` if you don't modify the chart?

thomasferrandiz (approved):

galal-hussein (approved):

brandond (approved):

PR #383, opened 2023-11-17
update cilium to v1.14.4

thomasferrandiz:


packages/rke2-cilium/generated-changes/patch/templates/cilium-configmap.yaml.patch
```diff
@@ -25,18 +25,18 @@
...
-@@ -434,7 +446,9 @@
-   tunnel-protocol: "vxlan"
- {{- end }}
- 
--{{- if eq .Values.tunnel "disabled" }}
+@@ -439,7 +451,9 @@
+ {{- else if .Values.routingMode }}
+   routing-mode: {{ .Values.routingMode | quote }}
+ {{- else }}
+-  {{- if eq .Values.tunnel "disabled" }}
```
manuelbuil, line 32: What are these changes about? Is there any bug that explains it?

thomasferrandiz: they were already present before but the file changed a bit in the new release so I had to adapt the patch

thomasferrandiz: and github's diff algorithm makes the change unreadable

manuelbuil (approved):

PR #386, opened 2023-12-14
add optional dhcp daemonset to rke2-multus

thomasferrandiz:
This PR adds an optional daemonSet to the multus chart.
It is required to use the dhcp ipam plugin with multus.

### DIFF
--- a/packages/rke2-multus/charts/templates/dhcp-daemonSet.yaml
+++ b/packages/rke2-multus/charts/templates/dhcp-daemonSet.yaml
@@ -1,9 +1,8 @@
----
 {{- if .Values.manifests.dhcpDaemonSet }}
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: multus-dhcp-ds
+  name: {{ .Release.Name }}-dhcp
   namespace: {{ .Release.Namespace }}
   labels:
 {{- include "multus.labels" . | indent 4 }}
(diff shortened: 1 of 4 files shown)


packages/rke2-multus/charts/templates/dhcp-daemonSet.yaml
```diff
@@ -0,0 +1,62 @@
+---
```
mgfritch, line 1:
```suggestion
```
remove?

packages/rke2-multus/charts/templates/dhcp-daemonSet.yaml
```diff
@@ -0,0 +1,62 @@
...
+{{- if .Values.manifests.dhcpDaemonSet }}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: multus-dhcp-ds
```
mgfritch, line 6:
```suggestion
  name: {{ .Release.Name }}-ds
```
(nit) add release name ?

brandond: I would also nit on putting -ds on the end; we know it's a daemonset. There's no need to use hungarian notation here.

thomasferrandiz: done!

manuelbuil (approved):

brandond (approved):

PR #395, opened 2024-01-15
rke2-whereabouts: fix gloablSystemRegistry and securityContext

thomasferrandiz:


packages/rke2-whereabouts/charts/templates/daemonset.yaml
```diff
@@ -42,9 +42,8 @@ spec:
               SLEEP=false /install-cni.sh &&
               /ip-control-loop -log-level debug
           securityContext:
-            # {{- toYaml .Values.securityContext | nindent 12 }}
-            privileged: true
```
brandond, line 46: does this actually work without being privileged? I am assuming it is privileged for a reason?

thomasferrandiz: it seems to work fine without privilege (with limited testing).
The old upstream chart had this same option to run the container without `privileged`.
In any case, the default in values.yaml is still to run whereabouts as privileged. 
I just wanted to let the option to the user if they need to remove it.

brandond (approved):

PR #399, opened 2024-01-24
Use windows values and fix mac bug

manuelbuil:
This PR:
* fixes a bug in flannel chart related to `macPrefix`
* updates values to use as default what vxlan windows interface is expecting (as flannel cni plugin will be mainly used for windows)

### DIFF
--- a/packages/rke2-flannel/generated-changes/patch/values.yaml.patch
+++ b/packages/rke2-flannel/generated-changes/patch/values.yaml.patch
@@ -26,7 +26,7 @@
    # flannel command arguments
    args:
    - "--ip-masq"
-@@ -25,28 +17,35 @@
+@@ -25,28 +17,36 @@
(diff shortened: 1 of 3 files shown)


thomasferrandiz (approved):

packages/rke2-flannel/generated-changes/patch/values.yaml.patch
```diff
@@ -31,7 +31,7 @@
    backend: "vxlan"
    # Port used by the backend 0 means default value (VXLAN: 8472, Wireguard: 51821, UDP: 8285)
 -  #backendPort: 0
-+  backendPort: 0
++  backendPort: 4789
```
rbrtbnfgl, line 34: Do we have to configure the port here? This could introduce issue if you use wireguard as backend

brandond (approved):

PR #416, opened 2024-03-04
Bump Vsphere CSI and CPI charts 

galal-hussein:


packages/rancher-vsphere-csi/generated-changes/overlay/templates/node/daemonset.yaml.orig
brandond, line 1: you appear to have a failed patch, or at least a file left over from a failed patch.

galal-hussein: yeah, I have tried multiple times, but it seems that make patch generates that file, let me try one more time

brandond: you'll need to fix the patch so that it applies cleanly, once that's done it should no longer generate the .rej/.orig files.

galal-hussein: I think I have fixed, for some reason the patch doesnt apply cleanly, I had to remove the generated changes dir and reapply the patches to the files

brandond (approved):

briandowns (approved):

PR #422, opened 2024-03-18
Allow IPTables mark mask to be configurable for canal

tkaepp:
Expose FELIX_IPTABLESMARKMASK to the canal charts to make it configurable. 

Related Issue: https://github.com/tailscale/tailscale/issues/591

> Calico takes the upper 16 bits of the netfilter packet mark for itself (aka 0xffff0000). This conflicts with Tailscale's use of 0x40000 and 0x80000, so trying to use Calico and Tailscale on the same Kubernetes machines will probably break Tailscale, or Calico, or both.

### DIFF
--- a/packages/rke2-canal/charts/values.yaml
+++ b/packages/rke2-canal/charts/values.yaml
@@ -99,6 +99,8 @@ calico:
   felixPrometheusMetricsEnabled: true
   # Disable XDP Acceleration as we do not support it with our ubi7 base image
   felixXDPEnabled: false
+  # Configure the mask that felix selects its iptables mark bits from
+  felixIptablesMarkMask: "0xffff0000"
   # Whether or not to masquerade traffic to destinations not within
   # the pod network.
   masquerade: true
--- a/packages/rke2-canal/charts/templates/daemonset.yaml
+++ b/packages/rke2-canal/charts/templates/daemonset.yaml
@@ -166,6 +166,8 @@ spec:
               value: {{ .Values.calico.felixFailsafeInboundHostPorts | quote }}
             - name: FELIX_FAILSAFEOUTBOUNDHOSTPORTS
               value: {{ .Values.calico.felixFailsafeOutboundHostPorts | quote }}
+            - name: FELIX_IPTABLESMARKMASK
+              value: {{ .Values.calico.felixIptablesMarkMask | quote }}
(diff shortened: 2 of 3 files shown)


manuelbuil:
Thanks for the PR. You need to increase this value to 01 so that the scripts generate a new tarball https://github.com/rancher/rke2-charts/blob/main-source/packages/rke2-canal/package.yaml#L2

packages/rke2-canal/charts/values.yaml
```diff
@@ -99,6 +99,8 @@ calico:
   felixPrometheusMetricsEnabled: true
   # Disable XDP Acceleration as we do not support it with our ubi7 base image
   felixXDPEnabled: false
+  # Configure the mask that felix selects its iptables mark bits from
+  felixIptablesMarkMask: 0xffff0000
```
brandond, line 103: Should this be quoted? YAML natively supports hex notation, so this is equivalent to `felixIptablesMarkMask: 4294901760` and I'm not sure if the env var will handle that properly, or if it is expected to be passed in as a hex value.

brandond (changes requested):

packages/rke2-canal/charts/values.yaml
tkaepp: I tried to edit the daemonset manully with `kubectl edit ds rke2-canal -n kube-system`. With quotes it works. Without I get the message:
> daemonsets.apps "rke2-canal" was not valid:
* patch: Invalid value: "map[spec:map[template:map[spec:map[]]]]": cannot convert int64 to string.

I tried it with both `0xff00ff00` and `4294901760`

Also inside the yaml `FELIX_IPTABLESREFRESHINTERVAL` is quoted and set to "60".

brandond: I mean quoted here in the values, not quoted in the template. It definitely needs to be, and already is, quoted in the template.

tkaepp: Yeah, you were right. The output with helm template without the quotes was:
```yaml
- name: FELIX_IPTABLESMARKMASK
  value: "4.29490176e+09"
```

Now it looks alright with the quotes.

brandond (approved):

rbrtbnfgl (approved):

PR #426, opened 2024-03-22
Add cni path config in whereabouts to make it work in k3s

manuelbuil:
The CNI binaries and CNI config files do not use the default paths in k3s, therefore, we need a way to configure those parameters if we want to use whereabouts

Multus version must be incremented because it embeds whereabouts

### DIFF
--- a/packages/rke2-whereabouts/charts/values.yaml
+++ b/packages/rke2-whereabouts/charts/values.yaml
@@ -46,5 +46,9 @@ tolerations:
 
 affinity: {}
 
+cniConf:
+  confDir: /etc/cni/net.d
+  binDir: /opt/cni/bin
+
 global:
   systemDefaultRegistry: ""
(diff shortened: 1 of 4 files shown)


packages/rke2-whereabouts/charts/values.yaml
```diff
@@ -46,5 +46,9 @@ tolerations:
 
 affinity: {}
 
+cni_conf:
```
brandond, line 49: does it have to be cni_conf? The convention is to use camelCase, not underscores. Maybe just `cni` or `cniConf`?
```suggestion
cni:
```

brandond (changes requested):

packages/rke2-whereabouts/charts/values.yaml
manuelbuil: good point! I just copied what Multus is using

brandond (approved):

mgfritch (approved):
(nit) doesn't affect the functionality since whereabouts is a subchart of multus, but should we also bump the whereabouts `packageVersion` too?

PR #429, opened 2024-03-27
Use the provided cni bin directory in the DHCP daemonset of Multus

manuelbuil:
The DHCP daemonset of multus is not reading the cni bin dir, which makes it unusable if the user changes it. This PR fixes that

### DIFF
--- a/packages/rke2-multus/charts/templates/daemonSet.yaml
+++ b/packages/rke2-multus/charts/templates/daemonSet.yaml
@@ -41,11 +41,10 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       nodeSelector: {{- toYaml .Values.labels.nodeSelector | nindent 8 }}
+      {{- with .Values.tolerations }}
       tolerations:
-      - operator: Exists
-        effect: NoSchedule
-      - operator: Exists
-        effect: NoExecute
+        {{- toYaml . | trim | nindent 8 }}
+      {{- end }}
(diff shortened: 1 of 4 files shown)


brandond (approved):

rbrtbnfgl (approved):

brandond:
While you are here @manuelbuil - it was noted on Rancher Users Slack that the multus DHCP daemonset does not honor the tolerations from the multus chart values. At the moment it has no tolerations whatsoever. You might look at adding that?

brandond:
This should probably be templated from chart values instead of hardcoded:
https://github.com/rancher/rke2-charts/blob/893bce41675211ac937c3416937f6fab48114db9/packages/rke2-multus/charts/templates/daemonSet.yaml#L43-L49
This needs to include the same `tolerations:` as the main daemonset:
https://github.com/rancher/rke2-charts/blob/893bce41675211ac937c3416937f6fab48114db9/packages/rke2-multus/charts/templates/dhcp-daemonSet.yaml#L27-L28

manuelbuil:
> While you are here @manuelbuil - it was noted on Rancher Users Slack that the multus DHCP daemonset does not honor the tolerations from the multus chart values. At the moment it has no tolerations whatsoever. You might look at adding that?

Thanks Brad, I was not aware of this. Let me add it to this PR too

brandond (approved):

dereknola (approved):

PR #441, opened 2024-04-12
Update Vsphere-CSI chart

galal-hussein:


packages/rancher-vsphere-csi/generated-changes/overlay/values.yaml.orig
brandond, line 1: delete the .orig

galal-hussein: shouldn't it be automatically cleaned with "make clean", or is that still broken?

brandond (approved):

packages/rancher-vsphere-csi/generated-changes/overlay/values.yaml.orig
brandond: apparently not lol

cwayne18 (approved):

PR #458, opened 2024-05-21
Fix updatecli scripts

rbrtbnfgl:
This PR adds few fixes for updatecli updates for the CNI:
- Adds `PATH` env to the updatecli scripts to fix an issue with `make prepare`
- Adds `set -e` to the updacli scripts to exit in case of errors and avoids to create a wrong update automated PR
- Sets `latest: true` to be sure to get the latest released version

### DIFF
--- a/updatecli/scripts/update-calico.sh
+++ b/updatecli/scripts/update-calico.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -eu
 if [ -n "$CALICO_VERSION" ]; then
 	current_calico_version=$(yq '.version' packages/rke2-calico/templates/crd-template/Chart.yaml)
 	if [ "$current_calico_version" != "$CALICO_VERSION" ]; then
--- a/updatecli/updatecli.d/updatecilium.yaml
+++ b/updatecli/updatecli.d/updatecilium.yaml
@@ -13,6 +13,7 @@ sources:
         release: true
         draft: false
         prerelease: false
+        latest: true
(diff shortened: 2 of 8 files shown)


updatecli/scripts/update-calico.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
```
manuelbuil, line 2: typically bash scripts also add "-eu" to make sure that all variables are set when they are called. It's a good practice to prevent unexpected executions

updatecli/updatecli.d/updatecilium.yaml
```diff
@@ -27,6 +28,7 @@ targets:
       environments:
         - name: CILIUM_VERSION
           value: '{{ source "cilium" }}'
+        - name: PATH
```
manuelbuil, line 31: don't you need to specify the PATH value? It just propagates it?

rbrtbnfgl: If not specified it uses the one already configured. The shell updatecli action will not propagate the environment variable by default

updatecli/scripts/update-calico.sh
rbrtbnfgl: I can add it to avoid possible errors on the script.

manuelbuil (approved):

thomasferrandiz (approved):

PR #473, opened 2024-06-07
Update configmap-nodelocal.yaml

letnab:
add dynamic to node-local-dns


brandond:
You can't just change files in the charts like that. The charts are all built from upstream using a set of build scripts. Please see the README in this repo; your PR will need to change the files at https://github.com/rancher/rke2-charts/blob/main-source/packages/rke2-coredns/generated-changes/overlay/templates/configmap-nodelocal.yaml

PR #474, opened 2024-06-07
add dynamic configmap-nodelocal

letnab:
add dynamic to configmap node-local-dns

### DIFF
--- a/packages/rke2-coredns/generated-changes/overlay/templates/configmap-nodelocal.yaml
+++ b/packages/rke2-coredns/generated-changes/overlay/templates/configmap-nodelocal.yaml
@@ -57,4 +57,14 @@ data:
         }
         prometheus :9253
         }
+    {{ range .Values.nodelocal.servers }}
+    {{- range $idx, $zone := .zones }}{{ $zone.zone }}{{ end -}}
+    {{- if .port }}:{{ .port }} {{ end -}}
+    {
+      {{- range .plugins }}
+        {{ .name }}
+          {{- end }}
+        bind {{ default $.Values.nodelocal.ip_address }} {{ template "nodelocalUpstreamDNSServerIP" }}
+    }
+    {{ end }}
 {{- end }}


brandond (changes requested):
You need to bump the packageVersion when changing the chart patches, if you are not also pulling in a new chart version from upstream:
https://github.com/rancher/rke2-charts/blob/0e59454f3f5dd15545c6649a1212c2c2953315e9/packages/rke2-coredns/package.yaml#L2C1-L2C19

packages/rke2-coredns/generated-changes/overlay/templates/configmap-nodelocal.yaml
```diff
@@ -57,4 +57,14 @@ data:
         }
         prometheus :9253
         }
+    {{ range .Values.nodelocal.servers }}
+    {{- range $idx, $zone := .zones }}{{ $zone.zone }}{{ end -}}
+    {{- if .port }}:{{ .port }} {{ end -}}
+    {
+      {{- range .plugins }}
+        {{ .name }}
+          {{- end }}
```
brandond, line 66: nit - shouldn't this match the indentation of the `{{- range .plugins }}` line?
```suggestion
      {{- end }}
```

brandond (changes requested):
Thanks for the contribution. One nit on the whitespace.

Can you please open an issue at https://github.com/rancher/rke2/issues describing your use case, and provide example chart values to allow our QA team to test this change?

packages/rke2-coredns/generated-changes/overlay/templates/configmap-nodelocal.yaml
letnab: yes. fixed/

brandond:
Can you show an example values that would use this? You should also patch the values.yaml to show where the new values to configure this should be set: https://github.com/rancher/rke2-charts/blob/ca72e83e2cfa6144f961fcdc631c1af0d7c3ad4e/charts/rke2-coredns/rke2-coredns/1.29.002/values.yaml#L388-L399

If the section shouldn't be present by default, add it commented out - for example:
https://github.com/rancher/rke2-charts/blob/ca72e83e2cfa6144f961fcdc631c1af0d7c3ad4e/charts/rke2-coredns/rke2-coredns/1.29.002/values.yaml#L374-L378

brandond:
I tried just copying the default `servers` value under the nodelocal section, and it does not appear to actually work:

```yaml
nodelocal:
  enabled: true
  servers:
  - plugins:
    - name: errors
    - configBlock: lameduck 5s
      name: health
    - name: ready
    - configBlock: |-
        pods insecure
        fallthrough in-addr.arpa ip6.arpa
        ttl 30
      name: kubernetes
      parameters: cluster.local in-addr.arpa ip6.arpa
    - name: prometheus
      parameters: 0.0.0.0:9153
    - name: forward
      parameters: . /etc/resolv.conf
    - name: cache
      parameters: 30
    - name: loop
    - name: reload
    - name: loadbalance
    port: 53
    zones:
    - zone: .
```

`Error: UPGRADE FAILED: template: rke2-coredns/templates/_helpers.tpl:260:5: executing "nodelocalUpstreamDNSServerIP" at <include "clusterDNSServerIP" .>: error calling include: template: rke2-coredns/templates/_helpers.tpl:239:36: executing "clusterDNSServerIP" at <.Values.global.clusterDNS>: invalid value; expected string`

PR #483, opened 2024-06-18
Enable nodelocal coredns with cilium replacing kube-proxy

thomasferrandiz:
This PR allows the user to configure coredns to use nodelocal cache with cilium and no kube-proxy.
It requires cilium Local Redirect Policy to be used.

Related issue: https://github.com/rancher/rke2/issues/5832

Ref: https://docs.cilium.io/en/v1.14/network/kubernetes/local-redirect-policy/#node-local-dns-cache

### DIFF
--- a/packages/rke2-coredns/generated-changes/overlay/templates/daemonset-nodelocal.yaml
+++ b/packages/rke2-coredns/generated-changes/overlay/templates/daemonset-nodelocal.yaml
@@ -29,7 +29,7 @@ spec:
       nodeSelector:
 {{ toYaml .Values.nodelocal.nodeSelector | indent 8 }}
 {{- end }}
-      hostNetwork: true
+      hostNetwork: {{not .Values.nodelocal.use_cilium_lrp }}
       dnsPolicy: Default  # Don't use cluster DNS.
       tolerations:
       - operator: Exists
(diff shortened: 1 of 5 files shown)


packages/rke2-coredns/generated-changes/overlay/templates/daemonset-nodelocal.yaml
```diff
@@ -61,15 +72,19 @@ spec:
...
             path: /health
             port: 8080
           initialDelaySeconds: 60
           timeoutSeconds: 5
         volumeMounts:
+{{- if .Values.nodelocal.use_cilium_lrp }}
         - mountPath: /run/xtables.lock
           name: xtables-lock
           readOnly: false
+{{- end}}
```
manuelbuil, line 87: By this change you are removing the `xtables.lock` when not using cilium_lrp, is that intentional? You'll be changing the current behaviour

thomasferrandiz: good catch, it should be the opposite

rbrtbnfgl:
Hi Thomas you need to change also this file if you are modifying the values file https://github.com/rancher/rke2-charts/blob/main-source/updatecli/scripts/cilium-values.yaml.patch.template it's used to generate the values patch file by updatecli.

thomasferrandiz:
> Hi Thomas you need to change also this file if you are modifying the values file https://github.com/rancher/rke2-charts/blob/main-source/updatecli/scripts/cilium-values.yaml.patch.template it's used to generate the values patch file by updatecli.

I'm not modifying anything in cilium actually only in coredns.

rbrtbnfgl:
Ok I read cilium on the PR title I thought the files were from cilium chart.

manuelbuil (approved):

PR #485, opened 2024-06-25
Upgrade calico image

galal-hussein:


rancher-max (approved):

packages/rke2-canal/package.yaml
```diff
@@ -1,2 +1,2 @@
 url: local
-packageVersion: 02
+packageVersion: 00
```
brandond, line 2: Should go to 03 I think?

galal-hussein: okay

dereknola (approved):

brandond (approved):

PR #486, opened 2024-06-26
feat: harvester-cloud-provider support k8s 1.30

FrankYang0529:
https://github.com/rancher/rancher/issues/45470


FrankYang0529:
Close the PR since we don't need it. Thanks.

PR #556, opened 2024-11-05
Level rke2-flannel CNI plugin config with k3s

Apsu:
This change follows how k3s configures flannel mostly, including adding the `bandwidth` plugin, except for leaving the cni version the same.

Most importantly and the main reason for this change, this adds `"forceAddress": true` to handle when the `cni0` interface, `/run/flannel/subnet.env`, and node's `podCIDR` don't match, without requiring an rke2 service restart/manual `cni0` intervention.

Reference: https://github.com/k3s-io/k3s/blob/master/pkg/agent/flannel/setup_linux.go

Issue: 
* https://github.com/rancher/rke2/issues/7244

### DIFF
--- a/packages/rke2-flannel/package.yaml
+++ b/packages/rke2-flannel/package.yaml
@@ -1,2 +1,2 @@
 url: https://github.com/flannel-io/flannel/releases/download/v0.26.0/flannel.tgz
-packageVersion: 01
+packageVersion: 02
--- a/packages/rke2-flannel/generated-changes/patch/templates/config.yaml.patch
+++ b/packages/rke2-flannel/generated-changes/patch/templates/config.yaml.patch
@@ -1,6 +1,25 @@
 --- charts-original/templates/config.yaml
 +++ charts/templates/config.yaml
-@@ -29,13 +29,13 @@
+@@ -16,6 +16,7 @@
+           "type": "flannel",
+           "delegate": {
+             "hairpinMode": true,
++            "forceAddress": true,
+             "isDefaultGateway": true
+           }
+         },
+@@ -24,18 +25,24 @@
+           "capabilities": {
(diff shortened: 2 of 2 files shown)


brandond (approved):

manuelbuil (approved):

manuelbuil:
@rbrtbnfgl any concern? Mine is only with rke2-windows but it does not use the chart option for the config, so it should not affect it

brandond:
lgtm but there appear to be issues with the packageversion

Apsu:
Happy to fix if I did the version bump wrong

packages/rke2-flannel/package.yaml
```diff
@@ -1,2 +1,2 @@
 url: https://github.com/flannel-io/flannel/releases/download/v0.26.0/flannel.tgz
-packageVersion: 00
+packageVersion: 01
```
manuelbuil, line 2: This should be 02. There was a 01 in the meanwhile

manuelbuil (changes requested):

packages/rke2-flannel/package.yaml
brandond: You need to rebase in addition to bumping the version

brandond (approved):

manuelbuil (approved):

manuelbuil:
@Apsu is it ok if we add this in the December release? We are already moving Flannel one minor version in the November release. I don't expect any problem but it's certainly a potential source of problems

Apsu:
Oh no problem at all. Thanks for being cautious!

brandond:
I created an issue in RKE2 to track missing bandwidth plugin support: https://github.com/rancher/rke2/issues/7244

PR #562, opened 2024-11-12
coredns: remove force_tcp option in nodelocaldns configmap

thomasferrandiz:
Ref: SURE-9325


rbrtbnfgl (approved):

manuelbuil:
Are there any potential bad consequences of removing that option?

thomasferrandiz:
> Are there any potential bad consequences of removing that option?

I don't think so, they remove it in the upstream manifest years ago and it's still the default.
The only thing I can think of is where a user has TCP port 53 open but not UDP:53.
But they can still adapt the configuration anyway.

manuelbuil (approved):

manuelbuil:
BTW, line 63 has another `force_tcp` line, should that also be removed?

thomasferrandiz:
> BTW, line 63 has another `force_tcp` line, should that also be removed?

I don't think so, the other `force_tcp` are still in the upstream manifest.

PR #633, opened 2025-02-20
Revert calico to `v3.29.1`

mgfritch:
Revert to calico `v3.29.1` on the canal chart

This reverts commit:
- https://github.com/rancher/rke2-charts/commit/e45fbbddb161b9d85bebc597d895ae3b0ca79937
- https://github.com/rancher/rke2-charts/commit/626f8fb5e57003eb14ac7f0c7d630dd57105d509

Bumps to latest flannel and calico images:
- rancher/hardened-flannel:v0.26.4-build20250218
- rancher/hardened-calico:v3.29.1-build20250220

### DIFF
--- a/packages/rke2-canal/charts/Chart.yaml
+++ b/packages/rke2-canal/charts/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: rke2-canal
 description: Install Canal Network Plugin.
-version: v3.29.1-build20250206
+version: v3.29.1-build20250220
 appVersion: v3.29.1
 home: https://www.projectcalico.org/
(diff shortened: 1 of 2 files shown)


brandond (changes requested):
Why? This doesn't un-publish v3.29.2, and v3.29.1 is still available from when it was previously published. Or are we trying to pull in a new build of an old version?

brandond (approved):
OK, I see we're not just reverting but also bumping to a new build of the v3.29.1 images with some base image fixes.

thomasferrandiz:
I'm confused don't we need to revert rke2-calico as well?

PR #670, opened 2025-04-21
Do not explicitly configure Calico MTU

bvsravikiran:
This change removes explicit configuration of Calico MTU to let Calico auto-detect host MTU and use it for workload interfaces.

For context, Calico auto-detects host network interface MTU and uses it to configure the MTU for workload interfaces [1][2]. When a non-zero MTU value is explicitly passed to Calico, it overrides the auto-detected MTU value. If the passed MTU value is greater than the host interface value, it could lead to fragmentation and packet loss.

[1] https://docs.tigera.io/calico/latest/networking/configuring/mtu#mtu-and-calico-defaults
[2] https://github.com/projectcalico/calico/blob/bda71ce13268ee63be485b0689ddeb5ff48f209f/manifests/canal.yaml#L59-L62

### DIFF
--- a/packages/rke2-canal/charts/values.yaml
+++ b/packages/rke2-canal/charts/values.yaml
@@ -107,7 +107,7 @@ calico:
   # Set Felix endpoint to host default action to ACCEPT.
   felixDefaultEndpointToHostAction: ACCEPT
   # Configure the MTU to use.
-  vethuMTU: 1450
+  vethuMTU: 0
   # Typha is disabled.
   typhaServiceName: none
   # Kubelet flex-volume-plugin-dir


brandond:
cc @manuelbuil

rbrtbnfgl:
I don't know if calico MTU autodetection will work fine with Canal that is using Flannel to forward the traffic between the pods.

bvsravikiran:
I have verified that both Calico and Flannel are auto-detecting the host MTU. While Calico monitors continuously with a background thread, Flannel does it once at startup. For example, when the host external network interface MTU is 9001, `/var/lib/calico/mtu` in calico-node container has 9001, and both the `flannel.1` interface the value of `FLANNEL_MTU` in `/run/flannel/subnet.env` in kube-flannel container is 8951. Similarly, when the host MTU is 1300, Flannel configured its MTU as 1250.

rbrtbnfgl:
Flannel considered the overhead required for VxLan header. I tried to configure 0 to get Calico autodetection and it seems that it's configuring the pod MTU to 1500 and the packets are fragmented on the tunnel.

PR #691, opened 2025-05-14
coredns: nodelocal should query the UDP service port

mgfritch:
The nodelocal `wait-coredns` container should query the coredns service port using UDP (instead of TCP)

See also: https://github.com/coredns/helm/commit/d579bfc7359dba42e9baa2f9b2dd3fc377b51f43
Issue: https://github.com/rancher/rke2/issues/8230


brandond:
good catch!

brandond:
Note that the upstream change to default to only UDP and not UDP+TCP may cause some unexpected changes if customer workloads expect DNS to be available over UDP; this may be something we want to revert in our chart packaging.

thomasferrandiz (approved):

rbrtbnfgl (approved):

PR #724, opened 2025-07-01
Add support for multus thick plugin

thomasferrandiz:
This PR adds the option to deploy using the thick plugin instead of the thin one.

It does not include multus-dynamic-networks-controller.


matttrach (approved):

vitorsavian (approved):

packages/rke2-multus/charts/values.yaml
```diff
@@ -139,3 +152,6 @@ global:
 rke2-whereabouts:
   enabled: false
   
+# When using the thick plugin, manifests.configMap should be set to true
```
brandond, line 155: Should or must? It looks like this PR changes the default to on, how does this affect users on upgrade?

packages/rke2-multus/charts/templates/daemonSet.yaml
```diff
@@ -59,8 +61,8 @@ spec:
           value: {{ .Values.cniplugins.skipcnis }}
       containers:
       - name: kube-{{ .Chart.Name }}
-        image: {{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}
-        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        image: {{ template "system_default_registry" . }}{{ .Values.images.thin.repository }}:{{ .Values.images.thin.tag }}
```
brandond, line 64: this will break users who are currently overriding the image via images.repository and images.tag. We need to continue to support the old structure (use Values.image if Values.images is nil) for backwards compatibility purposes.

brandond (changes requested):

packages/rke2-multus/charts/values.yaml
```diff
@@ -18,10 +18,14 @@
...
-  repository: rancher/hardened-multus-cni
-  tag: v4.2.1-build20250607
+images:
   pullPolicy: IfNotPresent
+  thin:
+    repository: rancher/hardened-multus-cni
+    tag: v4.2.1-build20250627
+  thick:
+    repository: rancher/hardened-multus-thick
+    tag: v4.2.1-build20250627
```
brandond, line 28: is there any possibility of building both of these into the same image and just running a different command? or is that too difficult?

packages/rke2-multus/charts/values.yaml
brandond: should we change the wording here to note that it must be true when using thick?

PR #775, opened 2025-09-09
Update traefik to v37.1.0 chart, use new hardened-traefik image.

dereknola:
Use separate traefik-crds chart that now exists (points to existing rke2-traefik-crd)

### DIFF
--- a/packages/rke2-traefik-crd/generated-changes/patch/Chart.yaml.patch
+++ b/packages/rke2-traefik-crd/generated-changes/patch/Chart.yaml.patch
@@ -1,0 +1,11 @@
+--- charts-original/Chart.yaml
++++ charts/Chart.yaml
+@@ -18,7 +18,7 @@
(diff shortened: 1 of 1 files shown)


packages/rke2-traefik-crd/package.yaml
```diff
@@ -0,0 +1,4 @@
+url: https://traefik.github.io/charts/traefik-crds/traefik-crds-1.11.0.tgz
```
brandond, line 1: I am concerned that we're going to have to patch the chart version or something here, otherwise helm will think that the chart is being downgraded from 34.2.002 -> 1.11.000

Maybe that's fine and it won't actually care, but we should test.

dereknola: Patched the chart name, hopefully the version works out.

brandond (approved):

manuelbuil (approved):

PR #917, opened 2026-03-10
Remove inconsistent READMEs from charts

manuelbuil:
JIRA: https://jira.suse.com/browse/SURE-11394

A customer is complaining because the README.MD is not aligned with our source of truth: values.yaml. That customer is using MD files (I guess via helm commands) to learn about the defaults.

This PR removes the READMEs to avoid confusion

### DIFF
--- a/packages/rke2-cilium/generated-changes/exclude/README.md
+++ b/packages/rke2-cilium/generated-changes/exclude/README.md
@@ -1,0 +1,1060 @@
+# cilium
+
+![Version: 1.19.3](https://img.shields.io/badge/Version-1.19.3-informational?style=flat-square) ![AppVersion: 1.19.3](https://img.shields.io/badge/AppVersion-1.19.3-informational?style=flat-square)
+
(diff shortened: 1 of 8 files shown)


manuelbuil:
We can actually remove them. I'll change the PR to do that instead

brandond (commented):
Do we need the full original content in the `exclude` dir to exclude it, or can we just have empty placeholder files there? If we don't need the original content we should just have an empty file, or maybe just a comment explaining why its excluded.

manuelbuil:
> Do we need the full original content in the `exclude` dir to exclude it, or can we just have empty placeholder files there? If we don't need the original content we should just have an empty file, or maybe just a comment explaining why its excluded.

Unfortunately, there is no way to provide "placeholder" such as an empty file. If we do that or add a comment, it creates a patch that again includes the whole content. I'll revert to removing the whole file and including it in "exclude"

brandond (approved):

thomasferrandiz (approved):

PR #918, opened 2026-03-10
Prepare for inclusion of rke2-security-responder

l-mb:


dereknola (approved):

brandond (changes requested):
I am confused why we would even install this if it's not enabled. The way to disable it would be by setting 
`disable: rke2-security-responder` in the RKE2 config as agreed upon [in the ADR](https://github.com/rancher/rke2/blob/master/docs/adrs/010-security-responder.md#configuration-interface-example). This will disable (and uninstall if it is already installed) the chart.

It seems like this is setting us up to install the chart, but have it just not install anything if it is not ALSO enabled by chart values? It doesn't make any sense to install the chart, but have the chart template render nothing if it's not enabled. Just, don't install it if you don't want it enabled.

l-mb:
> I am confused why we would even install this if it's not enabled. The way to disable it would be by setting `disable: rke2-security-responder` in the RKE2 config as agreed upon [in the ADR](https://github.com/rancher/rke2/blob/master/docs/adrs/010-security-responder.md#configuration-interface-example). This will disable (and uninstall if it is already installed) the chart.
> 
> It seems like this is setting us up to install the chart, but have it just not install anything if it is not ALSO enabled by chart values? It doesn't make any sense to install the chart, but have the chart template render nothing if it's not enabled. Just, don't install it if you don't want it enabled.

That was a somewhat clunky way to enable the "off-by-default" in the initial release. I've cleaned that up.

l-mb:
@brandond Does this address your feedback, or do I need to adjust something else?

(I suspect merging will only happen after the current release cycle, no worries.)

manuelbuil (approved):

brandond (approved):
LGTM!

caroline-suse-rancher (approved):
Thanks Lars!

PR #948, opened 2026-04-13
Split multus crd into its own sub-chart

thomasferrandiz:
and move whereabouts crds into this sub-chart.
We cannot split whereabouts crds into their own sub-chart
because whereabouts is itself a sub-chart of multus.

related issue:
- https://github.com/rancher/rke2/issues/8628

### DIFF
--- a/packages/rke2-multus/charts/Chart.yaml
+++ b/packages/rke2-multus/charts/Chart.yaml
@@ -1,4 +1,4 @@
-apiVersion: v2
+apiVersion: v1
 appVersion: 4.2.4
 dependencies:
 - condition: rke2-whereabouts.enabled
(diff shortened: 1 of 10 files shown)


brandond (approved):
We're going to need a transitional release with the CRDs in the old chart with the `keep` annotation, same as we're doing for gateway CRDs to get them out of traefik.

Yu-Jack:
Hi, I notice this PR from https://github.com/harvester/harvester/issues/10359. I have something to share about this CRD folder. If using a crd folder like this PR, when the content of CRD is updated, the existed CRD will not get updated because of this limitation https://helm.sh/docs/topics/charts/#limitations-on-crds from helm.

> CRDs are never reinstalled. If Helm determines that the CRDs in the crds/ directory are already present (regardless of version), Helm will not attempt to install or upgrade.

We hit this limitation before, so I just share this information here If my understanding is correct.

thomasferrandiz:
Adding annotations to CRDs: 
https://github.com/rancher/rke2-charts/pull/948

thomasferrandiz:
replaced by https://github.com/rancher/rke2-charts/pull/1116

PR #954, opened 2026-04-15
Bump to snapshot-controller v8.5.0

mgfritch:


thomasferrandiz (approved):

brandond (commented):
there are no changes to CRDs with this update?

mgfritch:
> there are no changes to CRDs with this update?

no, but I did forget to bump the crd appVersion. 

piraeusdatastore/helm-charts >= 5.0.0 does attempt to dynamically modify CRDs (via `installCRDs=true`) but we are not using that here.

brandond (approved):

PR #962, opened 2026-04-15
Fix image path in rke2-flannel chart

thomasferrandiz:


packages/rke2-flannel/generated-changes/patch/templates/daemonset.yaml.patch
```diff
@@ -13,7 +13,7 @@
 -        - /flannel
 -        - /opt/cni/bin/flannel
 +      - name: install-cni-plugins
-+        image: {{ template "system_default_registry" . }}{{ .Values.flannel.image_cni.repository }}:{{ .Values.flannel.image_cni.tag }}
++        image: {{ template "system_default_registry" . }}{{ .Values.flannel.flannel_cni.image.repository }}:{{ .Values.flannel.flannel_cni.image.tag }}
```
brandond, line 16: did this change in the upstream chart? This seems like it might be a breaking change.

brandond (approved):

packages/rke2-flannel/generated-changes/patch/templates/daemonset.yaml.patch
thomasferrandiz: yes it's in the upstream chart.
I had to do the update manually in #960 but I missed this change.

This time `helm template` works fine.

PR #973, opened 2026-05-05
Bundle ovirt csi driver

vitorsavian:
There is two pods in the chart that is `csi-resizer` and the `liveness-probe` that were added differently from what QA did since they are from the chart itself, the `csi-resizer` if the engine does no support resize they will be in CrashLoop and giving errors. 

So the question is -> should we remove the `csi-resizer`? we will lose this feature that the csi-driver has, but it will stop the errors or should we still give the chart and let users handle this?

Besides that, everything runs very smoothly, the only thing that might happen is that the image of the csi-driver is based on the latest and oracle apparently does no give anything besides `latest` or `test`

Issue -> https://github.com/rancher/rke2/issues/10304

### DIFF
--- a/packages/rke2-ovirt-csi-driver/generated-changes/patch/Chart.yaml.patch
+++ b/packages/rke2-ovirt-csi-driver/generated-changes/patch/Chart.yaml.patch
@@ -1,0 +1,12 @@
+--- charts-original/Chart.yaml
++++ charts/Chart.yaml
+@@ -1,8 +1,6 @@
+ apiVersion: v2
+-name: ovirt-csi-driver
++name: rke2-ovirt-csi-driver
+ description: Implementation of a CSI driver for oVirt.
+ type: application
+ version: 4.21.0
+ appVersion: "4.21.0"
+-icon: icons/kubernetes-logo.svg
+-kubeVersion: '>= 1.30.0 < 1.34.0'
--- a/packages/rke2-ovirt-csi-driver/generated-changes/patch/Chart.yaml.patch
+++ b/packages/rke2-ovirt-csi-driver/generated-changes/patch/Chart.yaml.patch
@@ -6,7 +6,9 @@
 +name: rke2-ovirt-csi-driver
  description: Implementation of a CSI driver for oVirt.
  type: application
- version: 4.21.0
- appVersion: "4.21.0"
+-version: 4.21.0
+-appVersion: "4.21.0"
 -icon: icons/kubernetes-logo.svg
 -kubeVersion: '>= 1.30.0 < 1.34.0'
++version: 4.15.0
++appVersion: "4.15.0"
--- a/packages/rke2-ovirt-csi-driver/generated-changes/patch/templates/csi_driver_controller.yaml.patch
+++ b/packages/rke2-ovirt-csi-driver/generated-changes/patch/templates/csi_driver_controller.yaml.patch
@@ -1,0 +1,11 @@
+--- charts-original/templates/csi_driver_controller.yaml
++++ charts/templates/csi_driver_controller.yaml
+@@ -37,7 +37,7 @@
+               apiVersion: v1
+               fieldPath: spec.nodeName
+         - name: OVIRT_CONFIG
+-          value: /tmp/config/ovirt-config.yaml
++          value: {{ quote .Values.csiController.ovirtController.env.ovirtConfig }}
+         - name: KUBERNETES_CLUSTER_DOMAIN
+           value: {{ quote .Values.kubernetesClusterDomain }}
+         image: {{ .Values.csiController.ovirtController.image.repository }}:{{ .Values.csiController.ovirtController.image.tag
--- a/packages/rke2-ovirt-csi-driver/generated-changes/patch/templates/csi_driver_controller.yaml.patch
+++ b/packages/rke2-ovirt-csi-driver/generated-changes/patch/templates/csi_driver_controller.yaml.patch
@@ -1,11 +1,8 @@
 --- charts-original/templates/csi_driver_controller.yaml
 +++ charts/templates/csi_driver_controller.yaml
-@@ -37,10 +37,10 @@
-               apiVersion: v1
-               fieldPath: spec.nodeName
+@@ -39,8 +39,8 @@
          - name: OVIRT_CONFIG
--          value: /tmp/config/ovirt-config.yaml
-+          value: {{ quote .Values.csiController.ovirtController.env.ovirtConfig }}
+           value: /tmp/config/ovirt-config.yaml
          - name: KUBERNETES_CLUSTER_DOMAIN
 -          value: {{ quote .Values.kubernetesClusterDomain }}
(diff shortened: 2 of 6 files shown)


packages/rke2-ovirt-csi-driver/generated-changes/patch/values.yaml.patch
```diff
@@ -0,0 +1,100 @@
+--- charts-original/values.yaml
++++ charts/values.yaml
+@@ -3,7 +3,7 @@
+     containerSecurityContext:
+       privileged: true
+     image:
+-      repository: olcne/csi-attacher
++      repository: registry.k8s.io/sig-storage/csi-attacher
```
brandond: We should have these already mirrored for harvester and vsphere; you'll need to replace these with the mirrored image, and either use tags we already have mirrored, or add the tags to artifact-mirror.

brandond (changes requested):
1. You need to patch in system-default-registry support to each of the images where they are referenced in templates. For examples see other charts such as:
  https://github.com/rancher/rke2-charts/blob/208e0384cbeeb50583d5d689bd5bb9994f8541b8/packages/rke2-cilium/generated-changes/patch/templates/_helpers.tpl.patch#L8-L14
  https://github.com/rancher/rke2-charts/blob/208e0384cbeeb50583d5d689bd5bb9994f8541b8/packages/rke2-cilium/generated-changes/patch/values.yaml.patch#L243-L244
  https://github.com/rancher/rke2-charts/blob/208e0384cbeeb50583d5d689bd5bb9994f8541b8/packages/rke2-cilium/generated-changes/patch/templates/cilium-agent/daemonset.yaml.patch#L39-L40
2. Need to add a value to enable resizer support, I'd probably slot it in under the existing `csiResizer` key as:
  ```yaml
    csiResizer:
      enabled: true
  ```

packages/rke2-ovirt-csi-driver/generated-changes/patch/values.yaml.patch
```diff
@@ -0,0 +1,111 @@
...
+-      repository: olcne/csi-provisioner
++      repository: registry.k8s.io/sig-storage/csi-provisioner
+       tag: v6.0.0
+     resources:
+       limits:
+@@ -26,10 +26,11 @@
+         cpu: 80m
+         memory: 64Mi
+   csiResizer:
++    enable: true
```
brandond: Should be `enabled` not `enable`, see for example: 
https://github.com/rancher/rke2-charts/blob/208e0384cbeeb50583d5d689bd5bb9994f8541b8/packages/rke2-calico/generated-changes/patch/values.yaml.patch#L43-L45

... and probably should be off by default, if it is not supported in our QA env? Lets have it off by default unless we hear otherwise.

```suggestion
+    enabled: false
```

brandond (changes requested):

packages/rke2-ovirt-csi-driver/generated-changes/patch/templates/csi_driver_controller.yaml.patch
```diff
@@ -0,0 +1,75 @@
+--- charts-original/templates/csi_driver_controller.yaml
++++ charts/templates/csi_driver_controller.yaml
+@@ -37,10 +37,10 @@
+               apiVersion: v1
+               fieldPath: spec.nodeName
+         - name: OVIRT_CONFIG
+-          value: /tmp/config/ovirt-config.yaml
++          value: {{ quote .Values.csiController.ovirtController.env.ovirtConfig }}
+         - name: KUBERNETES_CLUSTER_DOMAIN
+           value: {{ quote .Values.kubernetesClusterDomain }}
```
brandond: All instances of `.Values.kubernetesClusterDomain` should be modified prefer the value from `.Values.global.clusterDomain` if set. You can use `coalesce` for this, to take the first non-null value. See for example:
https://github.com/rancher/rke2-charts/blob/208e0384cbeeb50583d5d689bd5bb9994f8541b8/packages/rke2-coredns/generated-changes/patch/templates/configmap.yaml.patch#L8
```suggestion
           value: {{ coalesce .Values.global.clusterDomain .Values.kubernetesClusterDomain | quote }}
```

brandond (changes requested):

packages/rke2-ovirt-csi-driver/generated-changes/patch/templates/csi_driver_controller.yaml.patch
```diff
@@ -0,0 +1,85 @@
+--- charts-original/templates/csi_driver_controller.yaml
++++ charts/templates/csi_driver_controller.yaml
+@@ -37,10 +37,10 @@
+               apiVersion: v1
+               fieldPath: spec.nodeName
+         - name: OVIRT_CONFIG
+-          value: /tmp/config/ovirt-config.yaml
++          value: {{ quote .Values.csiController.ovirtController.env.ovirtConfig }}
```
brandond: I'm not sure this is actually used, or works if you change it.

I see /tmp/config as an empty tmpfs volume here: 
https://github.com/oracle-cne/catalog/blob/main/charts/ovirt-csi-driver-4.21.0/templates/csi_driver_controller.yaml#L178-L187

The file path is also hardcoded here:
https://github.com/oracle-cne/catalog/blob/main/charts/ovirt-csi-driver-4.21.0/templates/csi_driver_controller.yaml#L136-L143

If the user somehow created their own config file and DID want to change the path, there's no way to get it mounted into the pod.

The config file contains vars templated in from `.Values.ovirt.secretName`:
https://github.com/oracle-cne/catalog/blob/main/charts/ovirt-csi-driver-4.21.0/templates/csi_driver_controller.yaml#L135-L160

I think it's best to just leave it be and not change this. Creds go in the secret, the init container renders the secret into the config file at the expected, hardcoded path.

vitorsavian: It's very redundant, it will still use the ovirt-config.yaml in the init of the container. This is a left over of the config that I was first using to test the driver.

Let me clean this one.

brandond (approved):
lgtm!

packages/rke2-ovirt-csi-driver/generated-changes/patch/Chart.yaml.patch
```diff
@@ -0,0 +1,12 @@
+--- charts-original/Chart.yaml
++++ charts/Chart.yaml
+@@ -1,8 +1,6 @@
+ apiVersion: v2
+-name: ovirt-csi-driver
++name: rke2-ovirt-csi-driver
+ description: Implementation of a CSI driver for oVirt.
+ type: application
+ version: 4.21.0
+ appVersion: "4.21.0"
```
brandond: Should we set these to accurately reflect the version we're bundling?

```suggestion
 version: 4.15.0
 appVersion: "4.15.0"
```

brandond (approved):

PR #1002, opened 2026-05-21
Fixed calico toleration defined value

### LINKED ISSUE #10438: Calico taints are too permissive
**Environmental Info:**
RKE2 Version: v1.36.0

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

Single node etcd+controlplane+worker

**Describe the bug:**

When using calico v3.32.0, the tolerations are too permissive:
```
tolerations:
    - key: CriticalAddonsOnly
      operator: Exists
    - effect: NoSchedule
      operator: Exists
    - effect: NoExecute
      operator: Exists
```

This is what they were in v3.31.5:
```
tolerations:
    - effect: NoSchedule
      key: node-role.kubernetes.io/control-plane
      operator: Exists
    - effect: NoExecute
      key: node-role.kubernetes.io/etcd
      operator: Exists
    - effect: NoExecute
      key: node.kubernetes.io/not-ready
      operator: Exists
      tolerationSeconds: 300
    - effect: NoExecute
      key: node.kubernetes.io/unreachable
      operator: Exists
      tolerationSeconds: 300
```

When testing adding a second node to a single node all roles cluster with the intention of removing the first, the `calico-typha` pod is never evicted. When using CAPI, the machine controller will first drain (if enabled, is not in this case) detach volumes (if enabled, is not in this case), delete the infrastructure from the related hypervisor, and lastly perform the equivalent of a `kubectl delete node`. Since this pod can tolerate the node becoming unreachable, it is never rescheduled, however the `calico-node` on the new node fails to communicate with the service, which results in networking errors. Specifically, DNS resolution fails from the cluster-agent, so CAPI can no longer delete the node as the cluster-agent becomes disconnected.

**Steps To Reproduce:**

- Provision RKE2 from Rancher: v1.36.0+rke2r1
- Run Rancher's `Test_Provisioning_Single_Node_All_Roles_Drain` test with `SOME_K8S_VERSION=v1.36.0+rke2r1`

**Expected behavior:**

Calico is rescheduled

**Actual behavior:**

Calico is not rescheduled and cluster networking is broken forever

**Additional context / logs:**

Calico controller Pods:
```
NAME                                       READY   STATUS             RESTARTS         AGE   IP            NODE                                        NOMINATED NODE   READINESS GATES
calico-kube-controllers-84c466d665-cjmk9   1/1     Terminating        0                87m   10.44.2.200   test-single-node-drain-pool-0-m574r-dnbn7   <none>           <none>
calico-kube-controllers-84c466d665-nmtmc   0/1     CrashLoopBackOff   17 (4m29s ago)   76m   10.44.6.132   test-single-node-drain-pool-0-7mm2g-nwjmb   <none>           <none>

```

Tigera pod:
```
NAME                               READY   STATUS    RESTARTS   AGE   IP           NODE                                        NOMINATED NODE   READINESS GATES
tigera-operator-6b874999bb-kvpts   1/1     Running   0          87m   10.42.0.34   test-single-node-drain-pool-0-m574r-dnbn7   <none>           <none>

```

Typha Pod:
```
NAME                            READY   STATUS    RESTARTS   AGE   IP           NODE                                        NOMINATED NODE   READINESS GATES
calico-typha-599956d7f8-hmqjm   1/1     Running   0          87m   10.42.0.34   test-single-node-drain-pool-0-m574r-dnbn7   <none>           <none>
```

Service Definition:
```
NAME           TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE   SELECTOR
calico-typha   ClusterIP   10.45.162.177   <none>        5473/TCP   87m   k8s-app=calico-typha

```

rbrtbnfgl:
This fixes https://github.com/rancher/rke2/issues/10438

### DIFF
--- a/packages/rke2-calico-crd/package.yaml
+++ b/packages/rke2-calico-crd/package.yaml
@@ -1,2 +1,2 @@
 url: https://github.com/projectcalico/calico/releases/download/v3.32.0/crd.projectcalico.org.v1-v3.32.0.tgz
-packageVersion: 01
+packageVersion: 02
--- a/packages/rke2-calico/generated-changes/patch/values.yaml.patch
+++ b/packages/rke2-calico/generated-changes/patch/values.yaml.patch
@@ -1,20 +1,6 @@
 --- charts-original/values.yaml
 +++ charts/values.yaml
-@@ -9,6 +9,13 @@
- # on the operator.tigera.io/Installation API documented
- # here: https://docs.tigera.io/calico/latest/reference/installation/api#installationspec
- installation:
-+  controlPlaneTolerations:
-+  - key: "node-role.kubernetes.io/control-plane"
-+    operator: "Exists"
-+    effect: "NoSchedule"
-+  - key: "node-role.kubernetes.io/etcd"
-+    operator: "Exists"
-+    effect: "NoExecute"
-   enabled: true
-   # Uncomment to enable prometheus metrics reporting for node and typha.
-   # nodeMetricsPort: 9090
-@@ -16,7 +23,13 @@
+@@ -16,7 +16,13 @@
    # Path to the kubelet volume plugin directory used by the Calico CSI driver.
    # Set to "None" to disable the CSI driver. If unset, defaults to /var/lib/kubelet (CSI enabled).
    kubeletVolumePluginPath: "None"
@@ -29,7 +15,7 @@
    # --- General Image Settings ---
    # registry: <custom-registry>
    # imagePath: <optional-path>
-@@ -28,6 +41,7 @@
+@@ -28,6 +34,7 @@
    #
    # Example: --set installation.imagePullSecrets[0].name=my-existing-secret
    imagePullSecrets: []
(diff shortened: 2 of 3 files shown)


brandond:
It looked like there were also previously tolerationSeconds set on these, which I think is what was causing the pods to get evicted when the node was down. Do we need to re-add that as well?

cc @jakefhyde

jakefhyde:
> It looked like there were also previously tolerationSeconds set on these, which I think is what was causing the pods to get evicted when the node was down. Do we need to re-add that as well?
> 
> cc @jakefhyde

Yes, we were indeed relying on this to be evicted after 5 minutes (typically gives enough time for a cluster to be brought up).

brandond:
It appears that Kubernetes itself adds the not-ready and unreachable tolerations, but ONLY if the pod does not already tolerate all NoExec taints. So... this should work.

brandond (approved):

brandond:
For the record: https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#defaulttolerationseconds
> This admission controller sets the default forgiveness toleration for pods to tolerate the taints notready:NoExecute and unreachable:NoExecute based on the k8s-apiserver input parameters default-not-ready-toleration-seconds and default-unreachable-toleration-seconds if the pods don't already have toleration for taints node.kubernetes.io/not-ready:NoExecute or node.kubernetes.io/unreachable:NoExecute. The default value for default-not-ready-toleration-seconds and default-unreachable-toleration-seconds is 5 minutes.

Since upstream added these tolerations with no keys, the admission controller was not adding the default tolerations.

PR #1005, opened 2026-05-28
Bump Traefik to v3.7.1 and chart to v40.1.0

manuelbuil:
Bump Traefik to v3.7.1 and chart to v40.1.0


vitorsavian (approved):

brandond (approved):

manuelbuil:
I have just realized that we were not setting Traefik as defaultClass. I have changed that so that it matches what we had with ingress-nginx

PR #1012, opened 2026-06-03
Bump Traefik to v3.7.3

manuelbuil:
The ingressClassName change is what is needed to fix the failing rke2 test

### DIFF
--- a/packages/rke2-traefik/generated-changes/patch/values.yaml.patch
+++ b/packages/rke2-traefik/generated-changes/patch/values.yaml.patch
@@ -9,7 +9,7 @@
    # -- defaults to appVersion. It's used for version checking, even prefixed with experimental- or latest-.
    # To pin by digest, prefer `image.digest`. A `<version>@<digest>` combo is also accepted here; in that case the digest is what Kubernetes verifies and the version is informational (and can drift from the underlying image).
 -  tag:  # @schema type:[string, null]
-+  tag: v3.7.1-build20260601  # @schema type:[string, null]
++  tag: v3.7.3-build20260605  # @schema type:[string, null]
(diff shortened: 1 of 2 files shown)


packages/rke2-traefik/generated-changes/patch/values.yaml.patch
```diff
@@ -22,6 +22,15 @@
    # -- Number of pods of the deployment (only applies when kind == Deployment)
    replicas: 1
    # -- Number of old history to retain to allow rollback (If not set, default Kubernetes value is set to 10)
+@@ -124,7 +124,7 @@
+   # -- Create a default IngressClass for Traefik
+   enabled: true
+   isDefaultClass: true
+-  name: ""
++  name: "traefik"
```
brandond, line 32: Is this not a breaking change for users upgrading from previous releases who had set Traefik as default ingress?

Does this still respect our defaultIngressClass chart var?

manuelbuil: You are 100% right and I'm 100% idiot. I ran some tests with rke2-charts helm repo directly (no injection of global variables) and that confused me. I was seeing ingresses without defaultClass being ignored and thought this was a bug. Let me change this

manuelbuil: I left the values we had in 39.0.703 which work well: https://github.com/rancher/rke2-charts/blob/main/charts/rke2-traefik/rke2-traefik/39.0.703/values.yaml#L117-L121

manuelbuil:
Run tests and works

thomasferrandiz (approved):

rbrtbnfgl (approved):

PR #1013, opened 2026-06-03
Bump rke2-whereabouts to v0.9.4 and add ip-reconciler template

thomasferrandiz:


packages/rke2-whereabouts/charts/values.yaml
```diff
@@ -56,6 +56,16 @@ nodeSliceController:
   enabled: false
   podAnnotations: {}
 
+reconciler:
+  logLevel: debug
```
brandond: why debug by default? is this going to create a bunch of logs?

thomasferrandiz: it logs almost nothing even in debug mode but we can set it to info for clarity

rbrtbnfgl (approved):

PR #1047, opened 2026-07-03
Support kubernetesIngressNginx temporarily

manuelbuil:
Add a new function in _helpers called: "providers.kubernetesIngressNGINX.values" which coalesces .Values.providers.kubernetesIngressNGINX and .Values.providers.kubernetesIngressNginx

This way, we cover up for distracted users that are not aware of the Traefik breaking change in the new 40.x.x charts

## UPDATE ##
coalesce does not work well because the `.Values.providers.kubernetesIngressNginx` map does not include any default. As a consequence, we get errors such as:
```
Error: UPGRADE FAILED: template: rke2-traefik/templates/daemonset.yaml:58:15: executing "rke2-traefik/templates/daemonset.yaml" at <include "traefik.podTemplate" .>: error calling include: template: rke2-traefik/templates/_podtemplate.tpl:606:63: executing "traefik.podTemplate" at <.publishService.enabled>: nil pointer evaluating interface {}.enabled
```
One workaround is adding the old `kubernetesIngressNginx` map in values.yaml, so that the defaults are covered. The previous error is fixed but that is also not working beacause the new  kubernetesIngressNGINX map has new fields which are evaluated by some Traefik chart templates. As a result, when user uses the old map:
```
Error: UPGRADE FAILED: template: rke2-traefik/templates/daemonset.yaml:58:15: executing "rke2-traefik/templates/daemonset.yaml" at <include "traefik.podTemplate" .>: error calling include: template: rke2-traefik/templates/_podtemplate.tpl:609:26: executing "traefik.podTemplate" at <.modsec.enabled>: nil pointer evaluating interface {}.enabled
```
as "modsec" was not part of the old kubernetesIngressNginx.

Therefore I discarded `coalesce` and moved to `mergeOverwrite` which seems to work well. If  kubernetesIngressNginx is enabled, then we merge its values into kubernetesIngressNGINX map

### DIFF
--- a/packages/rke2-traefik/generated-changes/patch/templates/_helpers.tpl.patch
+++ b/packages/rke2-traefik/generated-changes/patch/templates/_helpers.tpl.patch
@@ -9,7 +9,44 @@
  {{- end -}}
  {{- end -}}
  
-@@ -486,3 +486,11 @@
+@@ -147,9 +147,20 @@
+ {{- print $servicePath | trimSuffix "-" -}}
+ {{- end -}}
+ 
++{{- define "providers.kubernetesIngressNGINX.values" -}}
++{{- $new := default dict .Values.providers.kubernetesIngressNGINX -}}
++{{- $old := default dict .Values.providers.kubernetesIngressNginx -}}
++{{- $provider := $new -}}
++{{- if and $old.enabled (not $new.enabled) -}}
++{{- $provider = mergeOverwrite (dict) $new $old -}}
++{{- end -}}
++{{- toYaml $provider }}
++{{- end -}}
++
+ {{- define "providers.kubernetesIngressNGINX.publishServicePath" -}}
+ {{- $defServiceName := printf "%s/%s" (include "traefik.namespace" .) (include "traefik.fullname" .) -}}
+-{{- $servicePath := default $defServiceName .Values.providers.kubernetesIngressNGINX.publishService.pathOverride }}
++{{- $provider := (include "providers.kubernetesIngressNGINX.values" . | fromYaml) -}}
++{{- $servicePath := default $defServiceName (dig "publishService" "pathOverride" "" $provider) }}
+ {{- print $servicePath | trimSuffix "-" -}}
+ {{- end -}}
+ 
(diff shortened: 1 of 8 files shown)


packages/rke2-traefik/generated-changes/patch/templates/_helpers.tpl.patch
```diff
@@ -9,7 +9,41 @@
...
-@@ -486,3 +486,11 @@
+@@ -147,9 +147,17 @@
+ {{- print $servicePath | trimSuffix "-" -}}
+ {{- end -}}
+ 
++{{- define "providers.kubernetesIngressNGINX.values" -}}
++{{- $new := default dict .Values.providers.kubernetesIngressNGINX -}}
++{{- $old := default dict .Values.providers.kubernetesIngressNginx -}}
++{{- $provider := coalesce (and $new.enabled $new) (and $old.enabled $old) $new -}}
++{{- toYaml $provider -}}
```
brandond: This really needs to be round tripped through toYaml and then back out via fromYaml everywhere we want to use it?

manuelbuil:
I need to test further, there is stuff not working well

packages/rke2-traefik/generated-changes/patch/templates/_helpers.tpl.patch
manuelbuil: I couldn't find a better way.

Without the `fromYaml` in the different templates where I am using the "include", the upgrade is failing because it returns a string:
```
Error: UPGRADE FAILED: template: rke2-traefik/templates/requirements.yaml:39:35: executing "rke2-traefik/templates/requirements.yaml" at <$kubernetesIngressNGINX.modsec>: can't evaluate field modsec in type string
```

I started with "print" instead of "toYaml" but it did not work (although no error). The templates were not reading the correct kubernetesIngressNGINX fields in values.yaml. I used AI to debug and saw that it returned a raw go map formatting. "toYaml" fixed it

brandond (approved):

matttrach (approved):

dereknola (approved):

PR #1099, opened 2026-07-17
rke2-ingress-nginx: Allow custom image to bypass system-default-registry when overriding backend ingress

dereknola:
https://github.com/rancher/rke2/issues/10832

Also fix the default backend image to one that available on both DockerHub and PRIME registries. 
Signed-off-by: Derek Nola <derek.nola@suse.com>

### DIFF
--- a/packages/rke2-ingress-nginx/generated-changes/patch/templates/_helpers.tpl.patch
+++ b/packages/rke2-ingress-nginx/generated-changes/patch/templates/_helpers.tpl.patch
@@ -1,21 +1,6 @@
 --- charts-original/templates/_helpers.tpl
 +++ charts/templates/_helpers.tpl
-@@ -70,11 +70,11 @@
- {{/*
- Get specific image
- */}}
--{{- define "ingress-nginx.image" -}}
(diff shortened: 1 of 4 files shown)


packages/rke2-ingress-nginx/generated-changes/patch/templates/_helpers.tpl.patch
```diff
@@ -15,16 +15,21 @@
...
      - name: modules
        mountPath: /modules_mount
  {{- end -}}
 +
-+{{- define "system_default_registry" -}}
-+{{- if .Values.global.systemDefaultRegistry -}}
-+{{- printf "%s/" .Values.global.systemDefaultRegistry -}}
-+{{- else -}}
++{{/*
++If the repository value is not FQIN, use the systemDefaultRegistry if defined.
```
brandond: Nit:
```suggestion
+If the repository value is not hostname or IP with at least one dot, or host:port, use the systemDefaultRegistry if defined.
```

vitorsavian (approved):

PR #1115, opened 2026-07-30
Update to cilium 1.20.0

### LINKED ISSUE #10973: Updatecli failed for cilium v1.20.0
Failed workflow run: https://github.com/rancher/rke2-charts/actions/runs/30479786744

ndreno:
Bumps the `rke2-cilium` chart from cilium 1.19.6 to [1.20.0](https://github.com/cilium/cilium/releases/tag/v1.20.0) and refreshes the generated patches against the new upstream chart.

Updatecli picked up 1.20.0 on 2026-07-29 but could not apply the patches, so this is the manual follow-up. Fixes rancher/rke2#10973.

## Changes

- `package.yaml`: url to `cilium-1.20.0.tgz`, `packageVersion` reset to `00`
- `values.yaml.patch` regenerated from `updatecli/scripts/cilium-values.yaml.patch.template` with the 1.20.0 image tags and digests. All 19 rke2 customizations carry over unchanged (portmap chaining, `cni.exclusive: false`, `ipam: kubernetes`, hubble disabled, envoy disabled, mirrored image repos, etcd toleration, `setNodeTaints: false`, portmap plugin, `global.systemDefaultRegistry`)
- All 12 template patches refreshed, plus the upstream README snapshot under `generated-changes/exclude/`
- `updatecli/scripts/cilium-values.yaml.patch.template` resynced with the regenerated `values.yaml.patch`, so the next automated run starts from matching context

### Beyond line-number refreshes

Two patch changes are not just context updates:

1. **`cilium-preflight/daemonset.yaml.patch`**: dropped the hunk for the `cilium-pre-flight-fqdn-precache` container, which was removed upstream in 1.20.0. This is the hunk that broke the automated run.

2. **`cilium-envoy/daemonset.yaml.patch`**: 1.20.0 adds `envoy.nodeLocality` along with a `lookup-bootstrap-locality` init container whose image reference did not go through `system_default_registry`. It is now wrapped like every other image in the chart. The feature is disabled by default, so nothing changes for existing installs, but enabling it on an airgapped cluster would otherwise have tried to pull from the upstream registry.

## Validation

- `make prepare` applies cleanly: no `.rej` and no `.orig` files
- `make patch` is idempotent against the committed `generated-changes`
- `helm template` renders successfully for IPv4-only, IPv6-only and dual-stack, with `enable-ipv4` / `enable-ipv6` resolving correctly from `global.clusterCIDRv4` / `global.clusterCIDRv6`
- Every rendered image is prefixed by `global.systemDefaultRegistry`, including the new envoy locality init container

## Before merging

The rancher mirrored images for 1.20.0 are not published yet. As of this writing the following return 404 on Docker Hub:

- `rancher/mirrored-cilium-cilium:v1.20.0`
- `rancher/mirrored-cilium-operator-generic:v1.20.0`
- `rancher/mirrored-cilium-cilium-envoy:v1.37.5-1782911245-7cffc778c923f68a77954a53b1a98d6b5353f004`
- `rancher/mirrored-cilium-certgen:v0.4.8`
- `rancher/mirrored-cilium-hubble-relay:v1.20.0`
- `rancher/mirrored-cilium-clustermesh-apiserver:v1.20.0`

`rancher/mirrored-cilium-hubble-ui` and `-hubble-ui-backend` at `v0.13.5` are already available. The mirrors need to land before this ships.

### DIFF
--- a/packages/rke2-cilium/generated-changes/exclude/README.md
+++ b/packages/rke2-cilium/generated-changes/exclude/README.md
@@ -1,12 +1,12 @@
 # cilium
 
-![Version: 1.19.6](https://img.shields.io/badge/Version-1.19.6-informational?style=flat-square) ![AppVersion: 1.19.6](https://img.shields.io/badge/AppVersion-1.19.6-informational?style=flat-square)
+![Version: 1.20.0](https://img.shields.io/badge/Version-1.20.0-informational?style=flat-square) ![AppVersion: 1.20.0](https://img.shields.io/badge/AppVersion-1.20.0-informational?style=flat-square)
 
 Cilium is open source software for providing and transparently securing
 network connectivity and loadbalancing between application workloads such as
 application containers or processes. Cilium operates at Layer 3/4 to provide
 traditional networking and security services as well as Layer 7 to protect and
-secure use of modern application protocols such as HTTP, gRPC and Kafka.
+secure use of modern application protocols such as HTTP and gRPC.
 
 A new Linux kernel technology called eBPF is at the foundation of Cilium.
 It supports dynamic insertion of eBPF bytecode into the Linux kernel at various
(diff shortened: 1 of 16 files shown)


matttrach (approved):

mgfritch (approved):
Thanks!

PR #1128, opened 2026-08-10
fix: keep netpol defaults in rke2-flannel values to prevent nil pointer on install

thomasferrandiz:
## Problem

The `packages/rke2-flannel/generated-changes/patch/values.yaml.patch` was deleting the entire `netpol` section from upstream flannel v0.28.9's `values.yaml`. However, the daemonset template still references `.Values.netpol.enabled`, causing:

```
nil pointer evaluating interface {}.enabled
```

when installing rke2-flannel v0.28.900.

## Fix

Remove the deletion lines for the `netpol` block from `values.yaml.patch`, so the upstream default (`netpol.enabled: false`) is preserved in the resulting `values.yaml`. The `global:` additions are retained and appended after `nodeSelector: {}` as before.

The resulting values.yaml will contain:
```yaml
netpol:
  enabled: false
  args:
  - "--hostname-override=$(MY_NODE_NAME)"
  - "--v=2"
  image:
    repository: registry.k8s.io/networking/kube-network-policies
    tag: v1.0.0
```

This satisfies the template reference to `.Values.netpol.enabled` and prevents the nil pointer error.

### DIFF
--- a/packages/rke2-flannel/generated-changes/patch/values.yaml.patch
+++ b/packages/rke2-flannel/generated-changes/patch/values.yaml.patch
@@ -64,7 +64,7 @@
          }
        ]
      }
-@@ -98,17 +92,19 @@
+@@ -98,9 +92,17 @@
        initialDelaySeconds: 5
        periodSeconds: 10
    tolerations:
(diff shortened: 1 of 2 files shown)


thomasferrandiz:
wrong fix

PR #1202, opened 2026-09-11
Update multus cni updatecli script

rbrtbnfgl:
- Update Multus cni updatecli script to update alse the CRD version and the chart version for release older then 1.36.
- Added the network controller image on the automation


mgfritch (approved):
awesome, thanks!

Resolves:

- incorrect multus-crd version during make charts
- missing update of the dynamic-network-controller image
- updatecli automation for the legacy (1.34-1.36) chart

PR #1205, opened 2026-09-16
Update Calico CRD to v3

rbrtbnfgl:
Update Calico CRD to `v3` to fix the API aggregation issue https://github.com/rancher/rke2/issues/11229


thomasferrandiz (approved):

mgfritch (approved):

rbrtbnfgl:
Before the update this requires documentation for the migration.

brandond (approved):
I am assuming that this will break things for anyone who has customized the configuration by hand, instead of doing it through the chart - since it is removing the old CRDs and adding new ones, anything currently stored in there but NOT managed by the chart or operator will be discarded.

Is there any value in allowing both to exist for a period of time?

rbrtbnfgl:
I'm moving it to draft because I did additional tests and I got some issues:
- As @brandond said if you update the chart version without a previous migration you lost all the defined resources with the old crd
- I tried the migration suggested by Calico and it works but after the update the crd installed previously with the migration process have conflicts with the ones installed with helm
- We probably need to keep both crd on the chart and do the migration with rke2?

rbrtbnfgl:
I changed it a bit. I added the crd v1 inside the crd alongside the v3 crd. I modified Calico chart that during the custom resource initialization it copies all the resources from v1 and redefines them as v3.

